PR #16, opened 2018-04-20
don't require store to be mutable to obtain writer for it

mykmelez:
`Store.write()` requires its `&self` argument to be mutable, but it isn't clear why, since `Store` is a "wrapper around an `lmdb::Database`," and it seems like those are intended to be long-lived and used by both read-only and read-write transactions, per [LMDB > Getting Started](http://www.lmdb.tech/doc/starting.html), which states:

> Generally databases should only be opened once, by the first transaction in the process. After the first transaction completes, the database handles can freely be used by all subsequent transactions.

Nor do I see anything about the Store abstraction that suggests that it needs to impose this additional constraint to support a more humane API. Although I'm new to this codebase (and Rust more generally), so I might be missing something!

Presumably it's better to reuse handles for perf, although it isn't clear how expensive it is to create them. In any case, here's a patch that removes the constraint. It passes tests (modulo warnings that a couple of Store instances no longer have to be mutable, so I've removed mutability from them).

### DIFF
--- a/src/readwrite.rs
+++ b/src/readwrite.rs
@@ -114,7 +114,7 @@ impl<K> Store<K> where K: AsRef<[u8]> {
         })
     }
 
-    pub fn write<'env>(&mut self, env: &'env Rkv) -> Result<Writer<'env, K>, lmdb::Error> {
+    pub fn write<'env>(&self, env: &'env Rkv) -> Result<Writer<'env, K>, lmdb::Error> {
         let tx = env.write()?;
         Ok(Writer {
             tx: tx,
--- a/src/readwrite.rs
+++ b/src/readwrite.rs
@@ -114,6 +114,8 @@ impl<K> Store<K> where K: AsRef<[u8]> {
         })
     }
 
+    /// Note: there may be only one write transaction active at any given time,
+    /// so this will block if any other writers currently exist for this store.
     pub fn write<'env>(&self, env: &'env Rkv) -> Result<Writer<'env, K>, lmdb::Error> {
         let tx = env.write()?;
         Ok(Writer {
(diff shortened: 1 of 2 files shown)


rnewman:
There are two reasons to make an argument `mut` in Rust:

1. It’s going to be mutated within scope, and so Rust requires that it be `mut`.
2. You want to use the borrow checker to ensure that only a single reference exists.

The typical use of the latter is to get help from the compiler to represent some external library (e.g., SQLite) constraints: a SQLite write transaction is exclusive, so even though the database handle isn’t mutated, we wish to act as if it were.

LMDB serializes writes: only a single write transaction can exist for a database, and attempts to open another will block. That’s only slightly different to SQLite: SQLite will raise `SQLITE_BUSY` instead. There are two ways to represent that in Rust:

1. With a non-mutable argument: a second attempt to begin a write transaction will block at runtime.
2. With a mutable argument: an attempt to begin a concurrent write transaction cannot be expressed in code.

I opted for the latter. The downside of that is that it’s also not possible to express a concurrent reader and writer, so this change probably makes sense…

src/readwrite.rs
```diff
@@ -114,7 +114,7 @@ impl<K> Store<K> where K: AsRef<[u8]> {
         })
     }
 
-    pub fn write<'env>(&mut self, env: &'env Rkv) -> Result<Writer<'env, K>, lmdb::Error> {
+    pub fn write<'env>(&self, env: &'env Rkv) -> Result<Writer<'env, K>, lmdb::Error> {
```
rnewman: It’s worth documenting right here that this will block if any other writers exist for this store.

mykmelez: Good idea, noted in 39d7e5d.

PR #18, opened 2018-04-20
add test of canonical read before write on same Writer

mykmelez:
With key-value stores, a common pattern is to read a value from the database for a given key, apply some business logic (potentially transforming the value), and then to write a new value to the database for that key.

When I try that with rkv, using the same Writer to first read and then write the value, I hit two errors that are unexpected:

```
error[E0502]: cannot borrow `writer` as mutable because it is also borrowed as immutable
   --> src/env.rs:254:9
    |
252 |         let _existing_value = writer.get("foo").expect("read");
    |                               ------ immutable borrow occurs here
253 |         // insert business logic here
254 |         writer.put("foo", &Value::I64(1234)).expect("wrote");
    |         ^^^^^^ mutable borrow occurs here
255 |         writer.commit().expect("committed");
256 |     }
    |     - immutable borrow ends here

error[E0505]: cannot move out of `writer` because it is borrowed
   --> src/env.rs:255:9
    |
252 |         let _existing_value = writer.get("foo").expect("read");
    |                               ------ borrow of `writer` occurs here
...
255 |         writer.commit().expect("committed");
    |         ^^^^^^ move out of `writer` occurs here
```

Perhaps my implementation is naive, or maybe rkv doesn't support its approach for good reason. Regardless, it'd be useful to include a test for the conventional way to do this.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -241,6 +241,17 @@ mod tests {
         }
     }
 
+    fn get_existing_foo(writer: &Writer<&str>) -> Option<i64> {
+        match writer.get("foo").expect("read") {
+            Some(Value::I64(num)) => Some(num),
+            _ => None,
+        }
+    }
+
+    fn set_foo(writer: &mut Writer<&str>, val: i64) -> Result<(), StoreError> {
+        writer.put("foo", &Value::I64(val))
+    }
+
     #[test]
     fn test_read_before_write() {
         let root = TempDir::new("test_read_before_write").expect("tempdir");
@@ -249,10 +260,12 @@ mod tests {
         let sk: Store<&str> = k.create_or_open("sk").expect("opened");
 
         let mut writer = sk.write(&k).expect("writer");
-        let _existing_value = writer.get("foo").expect("read");
-        // insert business logic here
-        writer.put("foo", &Value::I64(1234)).expect("wrote");
-        writer.commit().expect("committed");
+        let mut existing = get_existing_foo(&writer).unwrap_or(99);
+        existing += 1;
+        set_foo(&mut writer, existing).expect("success");
+        let updated = get_existing_foo(&writer).unwrap_or(99);
+        assert_eq!(updated, 100);
+        writer.commit().expect("commit");
     }
 
     #[test]
--- a/src/env.rs
+++ b/src/env.rs
@@ -242,6 +242,20 @@ mod tests {
     }
 
     #[test]
+    fn test_read_before_write() {
+        let root = TempDir::new("test_read_before_write").expect("tempdir");
+        fs::create_dir_all(root.path()).expect("dir created");
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        let sk: Store<&str> = k.create_or_open("sk").expect("opened");
+
+        let mut writer = sk.write(&k).expect("writer");
+        let _existing_value = writer.get("foo").expect("read");
+        // insert business logic here
+        writer.put("foo", &Value::I64(1234)).expect("wrote");
+        writer.commit().expect("committed");
+    }
+
+    #[test]
     fn test_concurrent_read_transactions_prohibited() {
         let root = TempDir::new("test_concurrent_reads_prohibited").expect("tempdir");
         fs::create_dir_all(root.path()).expect("dir created");
--- a/src/env.rs
+++ b/src/env.rs
@@ -241,28 +241,30 @@ mod tests {
         }
     }
 
-    fn get_existing_foo(writer: &Writer<&str>) -> Option<i64> {
-        match writer.get("foo").expect("read") {
-            Some(Value::I64(num)) => Some(num),
-            _ => None,
-        }
-    }
-
-    fn set_foo(writer: &mut Writer<&str>, val: i64) -> Result<(), StoreError> {
-        writer.put("foo", &Value::I64(val))
-    }
-
     #[test]
-    fn test_read_before_write() {
-        let root = TempDir::new("test_read_before_write").expect("tempdir");
+    fn test_read_before_write_num() {
+        let root = TempDir::new("test_read_before_write_num").expect("tempdir");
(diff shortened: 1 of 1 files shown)


rnewman:
You’ll see we do this here in Mentat:

https://github.com/mozilla/mentat/blob/master/tests/vocabulary.rs#L197

rnewman:
Oh, and to expand on the ‘artificial’ point: imagine you had this instead.

```rust
fn get_existing_foo(writer: &Writer) -> Option<i32> {
    writer.get(“foo”).map(value_to_my_i32)
}

fn set_foo(writer: &mut Writer, val: i32) -> Result<(), _> {
    writer.set(“foo”, val)
}

fn my_test() {
    let mut writer = …;
    let existing = get_existing_foo(&writer).unwrap_or(99);
    set_foo(&mut writer, existing).expect(“success”);
    writer.commit().expect(“commit”);
}
```

… the borrows for each function call only last for as long as the call does, and so by the time we come to `commit` there are no borrows of `writer`. This tends to be how real code is structured, and so in practice you rarely need to introduce additional scopes to constrain borrows.

mykmelez:
> This is something you mainly encounter in tests, because they’re long ‘scripts’ of calls without function boundaries. You need to manually introduce scopes to show that you’re done with something borrowed.
> 
> You need a pattern like this:
> 
> let existing_value;
> 
> // Scoped borrow of `writer`.
> {
>     existing_value = writer.get(“foo”).expect(“read”);
> }
> writer.commit().expect(“committed”);

This doesn't work, as *writer* remains immutably borrowed until the end of the function's scope (presumably for the life of *existing_value*).

> You can use the same pattern for the multiple-reference issue, or you can try to borrow mutably for reads, see if the compiler is any happier about that:
> 
> (&mut writer).read(…);

This causes the compiler to complain:

```
error[E0499]: cannot borrow `*writer` as mutable more than once at a time
   --> src/env.rs:255:9
    |
253 |         let _existing_value = (&mut writer).get("foo").expect("read");
    |                                     ------ first mutable borrow occurs here
254 |         // insert business logic here
255 |         writer.put("foo", &Value::I64(1234)).expect("wrote");
    |         ^^^^^^ second mutable borrow occurs here
256 |         writer.commit().expect("committed");
257 |     }
    |     - first borrow ends here
```


> … the borrows for each function call only last for as long as the call does, and so by the time we come to commit there are no borrows of writer. This tends to be how real code is structured, and so in practice you rarely need to introduce additional scopes to constrain borrows.

This works, and I've updated the test to use this approach.

Note that it's also possible to keep the code all in one function by destructuring the Value result via a `match` expression:

```
let mut existing = match writer.get("foo").expect("read") {
    Some(Value::I64(num)) => num,
    _ => 99,
};
```

Or its `if let` equivalent:

```
let mut existing = if let Some(Value::I64(num)) = writer.get("foo").expect("read") { num } else { 99 };
```

Unsure if that's preferable, however.

rnewman:
> This doesn't work, as writer remains immutably borrowed until the end of the function's scope (presumably for the life of `existing_value`).

Yes, this is a bit of a special case — the return value of `get` is scoped to a lifetime, because rkv takes pains to not copy unnecessarily. Your `match` example avoids this by silently copying `num`, so `existing` is of type `i64`, rather than of type `Value<'store>`.

rnewman:
The ergonomic option that I didn't get around to, because I hadn't had the opportunity to figure out how it felt, is:

```
impl<'a> Value<'a> {
    pub fn to_i64(self) -> Option<i64> {
        match self {
            Value::I64(val) -> Some(val),
            _ -> None,
        }
    }
}
```

and so on, or the more general conversion I hint at here:

```rust
// TODO: implement conversion between the two types of `Value` wrapper.
// This might be unnecessary: we'll probably jump straight to primitives.
enum OwnedValue {
    Bool(bool),
    U64(u64),
    I64(i64),
…
```

In this case you'd use something like:

```
let existing: Option<i64> = self.get("foo")?.map(|v| v.to_i64());
```

using `?` to bail on error, and `map` to turn `Option<Value<'a>>` into `Option<i64>`.

rnewman:
> Unsure if that's preferable, however.

And one final comment: I tend to encourage the use of `match`, because it's very easy to write convenience functions that clone. Something like:

```
match foo.do_something() {
    Ok(Some::Enum(With::Inner(_dontcare, Value::Thing(ref whatever)))) if whatever == mythang => {
        println!("Got it!");
    }
}
```

is super cheap, and much more convenient than a pile of chained transformations.

mykmelez:
If the value being retrieved is a Value::str, extracting its wrapped &str value is insufficient, as the &str is also scoped to the lifetime of the writer. So you have to copy the string:

```
let mut writer = sk.write(&k).expect("writer");
let mut existing = match writer.get("foo").expect("read") {
    Some(Value::Str(val)) => val,
    _ => "",
}.to_string();
existing.push('…');
writer.put("foo", &Value::Str(&existing)).expect("write");
```

In practice, you'll need to do that anyway, if you're retrieving it to modify it and then write it back. So it's a necessary allocation. It'd be useful to document, in any case.

I've added another test for this. Would it be better to convert these to examples in docs?

mykmelez:
> It’s fairer to say: you have to be done with the reference. You might parse it, or pass it to format!, or take a slice of it and turn that into a string, or compute the length. Or perhaps you don’t need the value at all.

Right, good point. Ok, I've added commentary to that effect into the two test functions (and refactored a bit to encapsulate the code better into those functions) in 3015951 (a rare commit ID abbreviation that is ambiguously hexadecimal due to no alpha digits!).

> Re doc examples: I think tests are fine for now. We will learn operations that ought to be added (perhaps we should have increment_and_get, or an Entry pattern, or map_set, if get-then-set is common in real code?).

Ok, that makes sense. I haven't added any documentation examples here, but I'll keep an eye out for such opportunities in the future.

mykmelez:
@rnewman Anything else I should do here, or is this good to land?

PR #19, opened 2018-04-25
implement deletion from store

mykmelez:
@rnewman: an API question here is whether to be consistent with the abbreviated method name [RwTransaction.del](https://docs.rs/lmdb/0.8.0/lmdb/struct.RwTransaction.html#method.del)—which is itself consistent with the underlying function name [mdb_del](https://docs.rs/lmdb-sys/0.8.0/lmdb_sys/fn.mdb_del.html)—or use the complete word *delete* as the name of the Writer method.

I've opted for the latter, as this is intended to be a humane interface, and humans usually speak in complete words. A minor consideration is that *del* may encourage [horizontal alignment](https://google.github.io/styleguide/javaguide.html#s4.6.3-horizontal-alignment) with consecutive *get* and *put* calls, which is arguably harmful (although also orthogonal and perhaps not something about which this library should have an opinion).

Regardless, the decision to call this *delete* was based on a weak opinion, weakly held, and I'm happy to rename it to *del* if you prefer.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -239,6 +239,51 @@ mod tests {
             assert_eq!(sk.get(r, "bar").expect("read"), Some(Value::Bool(true)));
             assert_eq!(sk.get(r, "baz").expect("read"), Some(Value::Str("héllo, yöu")));
         }
+
+        {
+            let mut writer = sk.write(&k).expect("writer");
+            writer.delete("foo").expect("deleted");
+            writer.delete("bar").expect("deleted");
+            writer.delete("baz").expect("deleted");
+            assert_eq!(writer.get("foo").expect("read"), None);
+            assert_eq!(writer.get("bar").expect("read"), None);
+            assert_eq!(writer.get("baz").expect("read"), None);
+
+            // Isolation. Reads still return values.
+            let r = &k.read().unwrap();
+            assert_eq!(sk.get(r, "foo").expect("read"), Some(Value::I64(1234)));
+            assert_eq!(sk.get(r, "bar").expect("read"), Some(Value::Bool(true)));
+            assert_eq!(sk.get(r, "baz").expect("read"), Some(Value::Str("héllo, yöu")));
+        }
+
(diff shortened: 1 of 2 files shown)


src/env.rs
```diff
@@ -239,6 +239,51 @@ mod tests {
...
+            writer.delete("bar").expect("deleted");
+            writer.delete("baz").expect("deleted");
+            assert_eq!(writer.get("foo").expect("read"), None);
+            assert_eq!(writer.get("bar").expect("read"), None);
+            assert_eq!(writer.get("baz").expect("read"), None);
+
+            writer.commit().expect("committed");
+        }
+
+        // Committed. Reads will succeed.
```
rnewman: "will succeed but return `None` to indicate a missing value." might be clearer.

src/readwrite.rs
```diff
@@ -71,6 +71,13 @@ impl<'env, K> Writer<'env, K> where K: AsRef<[u8]> {
             .map_err(StoreError::LmdbError)
     }
 
+    // TODO: duplicate data
```
rnewman: I think you're referring to `MDB_DUPSORT` support, yes? I think it's OK to make that a separate method, perhaps `remove_value`. You could even stub it out with `unimplemented!()`.

Even better would be to make this a method _only on a dupsort store_ — it would need a little bit of reorganizing of types and traits, but when I see "If the database does not support sorted duplicate data items (MDB_DUPSORT) the data parameter is ignored" in the docs, I see a footgun that we can avoid by using the type system.

src/env.rs
mykmelez: Got it, I've clarified in 2b0e0cf.

src/readwrite.rs
mykmelez: Yes, that's what I was referring to. I've replaced that comment in e973afb with an unimplemented!() function (and a test for it), with your "even better" variant as a code comment to prompt the implementer.

PR #21, opened 2018-05-15
make Manager consumable

mykmelez:
@rnewman `Manager::new` would have to be *pub* in order for a consumer to use it to create a Manager instance that manages database handles for them. But perhaps you have a different plan for how consumers should use Manager?

I suppose it somewhat begs the question to enable consumers to instantiate them directly, since a consumer can then evade the constraint that Manager imposes by instantiating it multiple times. One alternative would be to create a Manager singleton via a lazy static, i.e. something like:

```rust

#[macro_use]
extern crate lazy_static;

lazy_static! {
    #[derive(Debug)]
    static ref MANAGER: Manager = {
        Manager::new()
    };
}
```

Then keep `Manager::new()` private to force consumers to get or create Rkv instances via the singleton.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -23,7 +23,6 @@ use std::path::{
 
 use std::sync::{
     Arc,
-    Mutex,
     RwLock,
 };
 
@@ -35,14 +34,20 @@ use ::Rkv;
 
 /// A process is only permitted to have one open handle to each database. This manager
 /// exists to enforce that constraint: don't open databases directly.
+lazy_static! {
+    pub static ref MANAGER: RwLock<Manager> = {
+        RwLock::new(Manager::new())
+    };
+}
+
 pub struct Manager {
-    stores: Mutex<BTreeMap<PathBuf, Arc<RwLock<Rkv>>>>,
+    stores: BTreeMap<PathBuf, Arc<RwLock<Rkv>>>,
 }
 
 impl Manager {
-    pub fn new() -> Manager {
+    fn new() -> Manager {
         Manager {
-            stores: Mutex::new(Default::default()),
+            stores: Default::default(),
         }
     }
 
@@ -50,7 +55,7 @@ impl Manager {
     pub fn get<'p, P>(&self, path: P) -> Result<Option<Arc<RwLock<Rkv>>>, ::std::io::Error>
     where P: Into<&'p Path> {
         let canonical = path.into().canonicalize()?;
-        Ok(self.stores.lock().unwrap().get(&canonical).cloned())
+        Ok(self.stores.get(&canonical).cloned())
     }
 
     /// Return the open store at `path`, or create it by calling `f`.
@@ -58,8 +63,7 @@ impl Manager {
     where F: FnOnce(&Path) -> Result<Rkv, StoreError>,
           P: Into<&'p Path> {
         let canonical = path.into().canonicalize()?;
-        let mut map = self.stores.lock().unwrap();
-        Ok(match map.entry(canonical) {
+        Ok(match self.stores.entry(canonical) {
             Entry::Occupied(e) => e.get().clone(),
             Entry::Vacant(e) => {
                 let k = Arc::new(RwLock::new(f(e.key().as_path())?));
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -82,8 +82,7 @@ impl Manager {
     where F: FnOnce(&Path, c_uint) -> Result<Rkv, StoreError>,
           P: Into<&'p Path> {
         let canonical = path.into().canonicalize()?;
-        let mut map = self.stores.lock().unwrap();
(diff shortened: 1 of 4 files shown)


rnewman:
The idea was to not make the constructor public precisely to prevent any consumer from bypassing the single-instance restriction. If you’re brave enough to do that, just create stores directly rather than using `Manager`. There’s no benefit to having multiple managers.

tests/manager.rs
```diff
@@ -0,0 +1,32 @@
+extern crate rkv;
```
rnewman: License block.

src/manager.rs
```diff
@@ -40,7 +40,7 @@ pub struct Manager {
 }
 
 impl Manager {
-    fn new() -> Manager {
+    pub fn new() -> Manager {
```
rnewman: I don’t mind allowing `Manager` to be created for tests, but in that case please do so via a `#[cfg(test)]` `pub(crate)` method called `new_for_test` or something.

mykmelez:
> The idea was to not make the constructor public precisely to prevent any consumer from bypassing the single-instance restriction. If you’re brave enough to do that, just create stores directly rather than using Manager. There’s no benefit to having multiple managers.

Right, makes sense. However, in that case, how does one use Manager today? Perhaps I'm missing something, but there doesn't currently appear to be that single instance of Manager that consumers can access.

rnewman:
There isn’t, and some kind of thread-safe singleton must be implemented. If you can make it eager/on module load, that’s fine; it doesn’t really need to be very lazy.

tests/manager.rs
mykmelez: Fixed in 272c689.

mykmelez:
> There isn’t, and some kind of thread-safe singleton must be implemented. If you can make it eager/on module load, that’s fine; it doesn’t really need to be very lazy.

Making it eager fails on "calls in statics are limited to struct and enum constructors," and while we could replace Manager::new() with a direct call to the struct constructor, we'd still need to call RwLock::new() to make it threadsafe.

(In the past, we would've been able to use StaticRwLock, but that was deprecated in favor of lazy_static in https://github.com/rust-lang/rust/issues/27717.)

So 273ff58 uses lazy_static to add a thread-safe static singleton and makes `Manager::new()` private again.

mykmelez:
@fluffyemily and @ncalexan: per our conversation earlier today about expanding the reviewer corps for rkv, perhaps y'all can look at this?

mykmelez:
> This tree is all over the place, with reversions and a merge from master. Land just one commit with no merge, please.

I have mixed feelings about this, as I sometimes appreciate the "messy" history of a feature's development when digging at code archaeological sites. But, ok, I'll "squash and merge" when merging this pull request.

> Otherwise, this looks fine, but I have a slight preference for exposing Manager::singleton() or Manager::instance() rather than the all-caps very-special lazy static to consumers. Lazy static instances have strange types, or at least they did.

They still do, as the macro generates a type that dereferences to the specified type of the static (which appears to be of type *manager::MANAGER*). It isn't clear to me how likely it is for consumers to be confused by this type. Nevertheless, I agree that `Manager::singleton()` seems more idiomatic and ergonomic, despite its added length.

In 6293a29, I've made MANAGER private and added a `Manager::singleton()` static method that dereferences MANAGER from its macro-generated type to `RwLock<Manager>` and then returns a reference to it.

(An alternative would be for Manager to implement RwLock, such that consumers can call `Manager::read()` and `Manager::write()` directly. That would simplify the API but perhaps also obfuscate it.)

> In the diff from master, I noticed some tests were removed from src/env.rs. Explain why?

I didn't intend to do this, nor do I see it in this pull request, nor local diffs. Are you perhaps diffing locally against a local master branch that is behind upstream?

mykmelez:
> I have mixed feelings about this, as I sometimes appreciate the "messy" history of a feature's development when digging at code archaeological sites. But, ok, I'll "squash and merge" this pull request.

Erm, or did you mean to both squash all commits on this branch and also land it without a merge commit, fast-forwarding master to the squashed commit?

ncalexan:
> Erm, or did you mean to both squash all commits on this branch and also land it without a merge commit, fast-forwarding master to the squashed commit?

Squash so there's one commit; the false starts are rarely valuable.  Rebase that onto master.  Merge vs. fast-forward is your choice; I personally prefer merge with the r= on the merge commit, but GH makes that onerous.

mykmelez:
> Squash so there's one commit; the false starts are rarely valuable.

I'll agree to disagree. 😉 

> Rebase that onto master. Merge vs. fast-forward is your choice; I personally prefer merge with the r= on the merge commit, but GH makes that onerous.

Indeed, and it's quite useful to employ one of the flows that GitHub optimizes. In this case, it looks like "squash and merge" is the closest to what you prefer, so I'll do that.

PR #24, opened 2018-05-18
implement Value::Blob type (fixes #20)

### LINKED ISSUE #20: Blob support
Just about anything that we can get as `u8`s is something we can store…

This would be in incompatible version bump to add the type signature.

mykmelez:
Here's an implementation of Value::Blob along with tests of round-tripping both a [u8] and a [u16] that we convert to [u8] beforehand and convert back afterward. (Arguably the latter is more a test of the conversion functions than rkv itself, but it's useful documentation for developers who want to use rkv to store similar data.)

@fluffyemily Perhaps you can review this one?

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -143,7 +143,9 @@ impl Rkv {
 #[cfg(test)]
 mod tests {
     extern crate tempdir;
+    extern crate byteorder;
 
+    use self::byteorder::{ByteOrder, LittleEndian};
     use self::tempdir::TempDir;
     use std::fs;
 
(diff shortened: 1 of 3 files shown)


src/env.rs
```diff
@@ -404,6 +406,43 @@ mod tests {
...
+        fs::create_dir_all(root.path()).expect("dir created");
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        let sk: Store<&str> = k.create_or_open("sk").expect("opened");
+        let mut writer = sk.write(&k).expect("writer");
+
+        assert_eq!(writer.get("foo").expect("read"), None);
+        writer.put("foo", &Value::Blob(&[1, 2, 3, 4])).expect("wrote");
+        assert_eq!(writer.get("foo").expect("read"), Some(Value::Blob(&[1, 2, 3, 4])));
+
+        fn u16_to_u8(src: &[u16]) -> Vec<u8> {
```
fluffyemily: If this is something that users of rkv are going to have to do a lot, I'm a little confused as to why these functions are test only and not part of the API itself.

mykmelez: It's a good question. XULStore needs to store UTF-16 strings with [u16] backing storage, but I'm unsure at this point how common that'll be for rkv consumers in general. Thus my inclination is to await more evidence of usage before adding an abstraction for [u16] into rkv itself.

PR #25, opened 2018-05-29
implement iteration from start of store or specified key (fixes #23)

### LINKED ISSUE #23: support iterators and ranged lookups
In order to be able to iterate keys, and do so from an arbitrary point in the key space, rkv should expose LMDB's support for iterators and ranged lookups (behind humane abstractions as appropriate).

mykmelez:
This implements read-only iteration of `(key, value)` tuples from the start of the store or from a specified key, fixing #23 and providing a useful, humane subset of the functionality provided by LMDB cursors, which are abstracted away by this interface (one requests an _Iter_ directly from a _Reader_ without indirecting through a _Cursor_). Keys are `&[u8]`, while Values are _Value_ instances.

The _Reader.iter()_ and _Reader.iter_from()_ methods are analogous to the same-named methods of LMDB's [RoCursor](https://docs.rs/lmdb/0.8.0/lmdb/struct.RoCursor.html). An alternate interface would be a single _Reader.iter()_ method that accepts an _Option<AsRef<[u8]>_, which can be `None` (to iterate from the start of the store) or `Some(AsRef<[u8])` to start from that key. It wasn't clear which interface would be more ergonomic or idiomatic, however, and I defaulted to the one that is similar to LMDB's. I'm happy to reconsider (or consider a third alternative).

I discovered an issue in LMDB while developing this changeset and requested upstream pulls https://github.com/danburkert/lmdb-rs/pull/29 and https://github.com/danburkert/lmdb-rs/pull/30. They don't block landing this changeset, they'll just make it more humane by preventing a panic if you call _Reader.iter_from()_ on a key greater than any key in the store. We should update the minimum required version of LMDB once that lands upstream.

@rnewman I recall that you're around this week, so I'm requesting review from you, but please feel free to redirect to @ncalexan or @fluffyemily, both of whom have been reviewing in your stead, if you'd rather they review it.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -147,7 +147,7 @@ mod tests {
 
     use self::byteorder::{ByteOrder, LittleEndian};
     use self::tempdir::TempDir;
-    use std::fs;
+    use std::{fs, str};
 
     use super::*;
     use ::*;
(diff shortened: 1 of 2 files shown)


src/env.rs
```diff
@@ -456,4 +456,84 @@ mod tests {
...
+
+        // An iterator over an empty store returns no values.
+        {
+            let reader = sk.read(&k).unwrap();
+            let mut iter = reader.iter().unwrap();
+            assert!(iter.next().is_none());
+        }
+
+        let mut writer = sk.write(&k).expect("writer");
+        writer.put("foo", &Value::I64(1234)).expect("wrote");
```
rnewman: I would be interested in an addition to this test that includes non-ASCII key names.

src/env.rs
```diff
@@ -456,4 +456,84 @@ mod tests {
...
+        assert_eq!(val.expect("value"), Some(Value::F64(1234.0.into())));
+        assert!(iter.next().is_none());
+
+        // Iterators don't loop.  Once one returns None, additional calls
+        // to its next() method will always return None.
+        assert!(iter.next().is_none());
+
+        // Reader.iter_from() begins iteration at the first key equal to
+        // or greater than the given key.  In this case, the first (and only)
+        // key greater than "moo" is "noo", so that's the only result.
```
rnewman: Can you copypaste this test to include retrieval by prefix (e.g., `"no"`)?

src/readwrite.rs
```diff
@@ -104,6 +112,35 @@ impl<'env, K> Reader<'env, K> where K: AsRef<[u8]> {
     pub fn abort(self) {
         self.tx.abort();
     }
+
+    pub fn iter<'s>(&'s self) -> Result<Iter<'s>, StoreError> {
```
rnewman: This is actually equivalent to `iter_start` on `RoCursor`, so perhaps use that name for clarity?

src/env.rs
```diff
@@ -147,7 +147,7 @@ mod tests {
 
     use self::byteorder::{ByteOrder, LittleEndian};
     use self::tempdir::TempDir;
-    use std::fs;
+    use std::{fs, str};
```
rnewman: In Mentat we standardized on always using multi-line imports, always with a trailing comma, 'cos otherwise six months from now you'll end up reformatting all of these. Your choice whether to follow suit!

mykmelez:
> I'm noodling on what the impact is of eliding RoCursor.
> 
> Naturally we can't elide RwCursor, because it allows insertion and deletion during iteration, so there's an argument for keeping RoCursor.

Right. And RoCursor also allows retrieval of key/value pairs, which might be useful in conjunction with its cursor-positioning functionality (f.e. to retrieve a single key/value pair that is greater than or equal to a given key without having to instantiate an iterator only to discard subsequent pairs).

> There is some functionality on cursor that might be useful:
> 
>     Asking for the number of duplicates of the current key without traversing them.

Hmm, is that actually possible with a cursor? I haven't been able to find a way to do this.

>     Renewing (moving the cursor to a new read-only transaction) to avoid allocating.
>     Touching all pages in the cursor (prior to a write).
>     Arbitrary movement: backwards, to first or last, etc., without allocating a new cursor.

Indeed, all of these sound potentially useful!

> as well as a bunch of terrifying internals like moving nodes from one position in the tree to another by using two cursors as indicators.
> 
> I suspect the best thing to do is to expose a helpful iterator interface as you've done here, but bear in mind that we might subsequently find a use case that requires exposing cursors more directly.

That seems reasonable to me. And we might do that by simply exposing the "lmdb" crate's handles (as the lmdb crate itself does with LMDB handles, f.e. via the `lmdb::Cursor.cursor()` function, which returns a `*mut ffi::MDB_cursor`).

However, we might instead build an abstraction over LMDB cursors that provides a humane API (f.e. one that doesn't require calling `lmdb::Cursor.get()` with an *MDB_cursor_op* to traverse the store). I'll leave that decision as an exercise to a future hacker!

src/env.rs
mykmelez: I've added this in 4e33519, which tests keys with both Latin-1 accented and Chinese characters.

src/env.rs
mykmelez: Yes, done in fba629f.

src/readwrite.rs
mykmelez: Good point! I've done so in 7a0d338.

rnewman:
> Hmm, is that actually possible with a cursor? I haven't been able to find a way to do this.

`mdb_cursor_count` does it.

http://www.lmdb.tech/doc/group__internal.html#gafb422278c217a4d3ff0cccda47048006

I don't know if that's exposed via lmdb_rs.

mykmelez:
> mdb_cursor_count does it.

Ah, right. It isn't explicitly exposed, but since lmdb-rs Cursor allows you to retrieve a pointer to the underlying MDB_cursor, and lmdb-rs's FFI is public, a consumer with access to the lmdb-rs Cursor object could retrieve an MDB_cursor pointer and call mdb_cursor_count() on it themselves. Ditto for other cursor operations in LMDB that aren't explicitly exposed by lmdb-rs.

And regardless, your point is that there are operations on cursors that are useful, and we might consider exposing them in the future (regardless of what lmdb-rs exposes explicitly) in order to enable consumers to take advantage of those operations. I totally agree!

PR #26, opened 2018-05-29
update version for release to crates.io

mykmelez:
We should publish a new release, as we've landed several features recently, including the Manager, iteration, and Value::Blob, the latter of which deserves a version bump per @rnewman's comment in #20.


mykmelez:
Tagged as https://github.com/mozilla-prototypes/rkv/tree/0.3.0 and published to https://crates.io/crates/rkv.

PR #28, opened 2018-06-05
use consistent terminology in Manager implementation and comments

mykmelez:
@ncalexan There's some inconsistency in the terminology used by manager.rs to describe the things it manages.

The Rkv struct wraps an LMDB "environment," while the Store struct wraps an LMDB "database," and manager.rs manages the former, but it sometimes describes them as stores rather than environments. This change ensures we use the terms "environment" and "store" consistently in that module.

Note that I asked @rnewman a related question last month:

>Something that has been puzzling me about LMDB recently is that http://www.lmdb.tech/doc/starting.html says, “Because of this, do not mdb_env_open() a file multiple times from a single process. Instead, share the LMDB environment that has opened the file across all threads.” And rkv::Manager enforces this constraint…
>
>But other references, like http://www.lmdb.tech/doc/, say, “Do not have open an LMDB database twice in the same process at the same time. Not even from a plain open() call - close()ing it breaks flock() advisory locking.” And database != environment in LMDB.
>
>And one can open multiple (named) databases from a single environment. So I’ve been wondering if that second statement uses the word “database” colloquially, to refer to an LMDB environment; or if it’s actually the case that one shouldn’t open the same LMDB database (a.k.a. Rkv::Store) twice in the same process, in which case Rkv should enforce that constraint.

And he responded:

> I believe both constraints apply. The DB one is most practically relevant.

So we may want manager.rs to manage databases too, and not just environments. But that isn't what it does today, and we should be clear about that. Hence this change.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -36,8 +36,8 @@ use error::{
 
 use ::Rkv;
 
-/// A process is only permitted to have one open handle to each database. This manager
-/// exists to enforce that constraint: don't open databases directly.
+/// A process is only permitted to have one open handle to each Rkv environment.
+/// This manager exists to enforce that constraint: don't open environments directly.
 lazy_static! {
     static ref MANAGER: RwLock<Manager> = {
         RwLock::new(Manager::new())
@@ -45,13 +45,13 @@ lazy_static! {
 }
 
 pub struct Manager {
-    stores: BTreeMap<PathBuf, Arc<RwLock<Rkv>>>,
+    environments: BTreeMap<PathBuf, Arc<RwLock<Rkv>>>,
 }
 
 impl Manager {
     fn new() -> Manager {
         Manager {
-            stores: Default::default(),
+            environments: Default::default(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


mykmelez:
> If we really aren't supposed to open an LMDB database (== Rkv store) more than once in a process, will the manager grow to hand out Arc<Store> and ensure that the Store itself is Sync?

I think so, yes. @rnewman should chime in if he was thinking otherwise!

mykmelez:
>> If we really aren't supposed to open an LMDB database (== Rkv store) more than once in a process, will the manager grow to hand out Arc and ensure that the Store itself is Sync?
> 
> I think so, yes. @rnewman should chime in if he was thinking otherwise!

@ncalexan I filed this as #29.

rnewman:
The LMDB database handle is `Copy`; that is, one doesn't need to pass around `Arc`. The restriction is that a file path is only opened once, not that only one instance of the handle exists in memory. Now if `Store` is heavyweight, or you need a good place to handle close semantics, then you might want borrowing, but it seems to be easier to pass by value.

PR #30, opened 2018-06-06
replace the deprecated tempdir crate with tempfile

ncloudioj:
Per [this note](https://github.com/rust-lang-nursery/tempdir#deprecation-note), tempdir has been merged into [tempfile](https://github.com/Stebalien/tempfile). Fixing it as my good-first-bug to rkv ;)

r? @mykmelez

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,4 +30,4 @@ features = ["derive"]
 
 [dev-dependencies]
 byteorder = "1"
-tempdir = "0.3"
+tempfile = "3"
--- a/src/env.rs
+++ b/src/env.rs
@@ -142,7 +142,7 @@ impl Rkv {
 
 #[cfg(test)]
 mod tests {
-    extern crate tempdir;
+    extern crate tempfile;
     extern crate byteorder;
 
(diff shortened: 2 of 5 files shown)


mykmelez:
Nice catch!

I wish there was a way for crates.io to flag deprecated modules (and, ideally, direct you to their replacements!). I suppose it isn't theoretically necessary to update a crate that works fine for your current use cases, although it would be helpful not to have to scramble to switch crates (as opposed to merely scrambling to upgrade an existing crate) when a security vulnerability or other significant issue is discovered.

ncloudioj:
Thanks @mykmelez 

Indeed, It'll be nice to have some support from crates.io for this, I happened to see this deprecation note when I was poking around some other stuff in the codebase.

In terms of keeping dependencies in sync with Gecko, yes, I agree with you that we should be avoiding unnecessary footprint from dependencies.

Next, as I am getting more familiar with rkv as well as lmdb-rs, I guess I will be working on adding some examples in hope that those could help folks get up-to-speed when checking out rkv.

PR #32, opened 2018-06-07
add rustfmt

ncloudioj:
Formatted by rustfmt, only added one option to rustfmt.toml to favor the current [import indent](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#vertical-1).

Although, it still flattened the single import from `use::foo::{Bar}` to `use::foo::Bar` :/

### DIFF
--- a/src/integer.rs
+++ b/src/integer.rs
@@ -8,13 +8,11 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-use std::marker::{
-    PhantomData,
-};
+use std::marker::PhantomData;
 
 use bincode::{
-    Infinite,
     serialize,
+    Infinite,
 };
 
 use lmdb::{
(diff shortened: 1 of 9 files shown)


src/integer.rs
```diff
@@ -52,7 +45,10 @@ pub trait PrimitiveInt: EncodableKey {}
 
 impl PrimitiveInt for u32 {}
 
-impl<T> EncodableKey for T where T: Serialize {
+impl<T> EncodableKey for T
+where
+    T: Serialize,
+{
```
mykmelez: In general, I like the way rustfmt formats *where* clauses, and it's reasonable to do that consistently, although I might make an exception for very simple declarations like this one, if it were possible to do so (which it doesn't appear to be).

src/manager.rs
```diff
@@ -30,18 +24,14 @@ use std::sync::{
...
-use ::Rkv;
+use Rkv;
 
 /// A process is only permitted to have one open handle to each Rkv environment.
 /// This manager exists to enforce that constraint: don't open environments directly.
 lazy_static! {
-    static ref MANAGER: RwLock<Manager> = {
-        RwLock::new(Manager::new())
-    };
+    static ref MANAGER: RwLock<Manager> = { RwLock::new(Manager::new()) };
```
mykmelez: This is fair enough, although it would be even simpler and more readable to also remove the brackets:

```rust
    static ref MANAGER: RwLock<Manager> = RwLock::new(Manager::new());
```

Unsure if that's configurable, but it isn't a big deal. I'm ok either way.

ncloudioj:
Closing it via #40.

PR #35, opened 2018-06-08
add an example for iterators

ncloudioj:
r? @mykmelez

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,3 +7,5 @@ cargo run --example $name
 ```
 
 * [`simple-store`](simple-store.rs) - a simple key/value store that showcases the basic usage of rkv.
+
+* [`iterator`](iterator.rs) - a demo that showcases the basic usage of iterators in rkv.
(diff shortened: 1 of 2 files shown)


ncloudioj:
Noticed that the iterator shares the same type as `Reader::get` for the return values (`Result<Option<Value>>`). 

Shall we consider simplifying it for the iterators? For instance, the value was wrapped as `Ok(Some(Str("foo")))` in this demo after unwrapping the outmost Result. It makes sense to have Option and Result for `Reader::get`, whereas the `Result<Option<` seems somewhat redundant here.

mykmelez:
> Shall we consider simplifying it for the iterators? For instance, the value was wrapped as Ok(Some(Str("foo"))) in this demo after unwrapping the outmost Result. It makes sense to have Option and Result for Reader::get, whereas the Result<Option< seems somewhat redundant here.

Good point. At least the Result seems unnecessary, by my reading of [Lmdb::Iter::next()](https://github.com/danburkert/lmdb-rs/blob/master/src/cursor.rs#L241-L260), which always returns an `Option`.

PR #38, opened 2018-06-13
workaround unc path on windows

ncloudioj:
Workaround the UNC path on Windows, all the tests passed in the current stable toolchain "stable-x86_64-pc-windows-msvc"

Note that:
* It can't handle the paths longer than MAX_PATH
* It can't handle the paths with the non-utf8 characters

@mykmelez

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -12,6 +12,12 @@ use std::collections::{
     BTreeMap,
 };
 
+use std::io::{
+    self,
+    Error,
+    ErrorKind,
+};
+
 use std::collections::btree_map::{
     Entry,
 };
@@ -30,6 +36,8 @@ use std::sync::{
     RwLock,
 };
 
+use url::Url;
+
 use error::{
     StoreError,
 };
@@ -42,6 +50,20 @@ lazy_static! {
     static ref MANAGER: RwLock<Manager> = {
         RwLock::new(Manager::new())
     };
+}
+
+// Workaround the UNC path on Windows, see https://github.com/rust-lang/rust/issues/42869.
+// Otherwise, `Env::from_env()` will panic with error_no(123).
+fn canonicalize_path<'p, P>(path: P) -> io::Result<PathBuf>
+where P: Into<&'p Path> {
+    let canonical = path.into().canonicalize()?;
+    if cfg!(target_os = "windows") {
+        let url = Url::from_file_path(&canonical)
+            .map_err(|_e| Error::new(ErrorKind::Other, "URL passing error"))?;
+        return url.to_file_path()
+            .map_err(|_e| Error::new(ErrorKind::Other, "path canonicalization error"));
+    }
+    Ok(canonical)
 }
 
 pub struct Manager {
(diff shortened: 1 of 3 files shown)


src/manager.rs
```diff
@@ -44,6 +46,24 @@ lazy_static! {
     };
 }
 
+// Workaround the UNC path on Windows, see https://github.com/rust-lang/rust/issues/42869.
+// Otherwise, `Env::from_env()` will panic with error_no(123).
+fn canonicalize_path<'p, P>(path: P) -> io::Result<PathBuf>
+where P: Into<&'p Path> {
+    let canonical = path.into().canonicalize()?;
+    {
```
mykmelez: It seems like we should preface this block to compile it only on Windows:

```rust
    #[cfg(windows)]
```

Then other platforms won't pay the cost of the workaround.

src/manager.rs
```diff
@@ -44,6 +46,24 @@ lazy_static! {
     };
 }
 
+// Workaround the UNC path on Windows, see https://github.com/rust-lang/rust/issues/42869.
+// Otherwise, `Env::from_env()` will panic with error_no(123).
```
mykmelez: It'd be good to file this upstream, in [lmdb-rs](https://github.com/danburkert/lmdb-rs), or perhaps farther upstream, in LMDB itself, although the latter requires grokking the [OpenLDAP Issue Tracking System](https://www.openldap.org/its/index.cgi), so I'd probably start with the former. 😉

src/manager.rs
```diff
@@ -44,6 +46,24 @@ lazy_static! {
...
+where P: Into<&'p Path> {
+    let canonical = path.into().canonicalize()?;
+    {
+        // TODO: use OsStr instead of Str in case that there are non-utf8 characters
+        // in the path. Pattern matching for OsStr will be implemented by
+        // https://github.com/rust-lang/rust/issues/49802
+        let mut canonical_path = canonical.to_str().unwrap();
+        if canonical_path.starts_with(r"\\?\") {
+            canonical_path = &canonical_path[r"\\?\".len()..];
+            return Ok(PathBuf::from(canonical_path));
```
mykmelez: Hmm, I wonder if it would be more robust to unconditionally strip the UNC prefix from paths via the [Url crate](https://doc.servo.org/url/struct.Url.html), i.e. something like:

```rust
Url::from_file_path(canonical).expect("URL").to_file_path().expect("path")
```

I don't know if this would cover more cases or fail less often than the current implementation, however. What do you think?

src/manager.rs
```diff
@@ -44,6 +46,24 @@ lazy_static! {
     };
 }
 
+// Workaround the UNC path on Windows, see https://github.com/rust-lang/rust/issues/42869.
```
mykmelez: Ah, https://github.com/rust-lang/rust/issues/42869, I [know it well](https://github.com/alexcrichton/cargo-vendor/pull/71#issuecomment-387570072) (unfortunately).

src/manager.rs
ncloudioj: Good point, will tweak it a bit based on all other review feedback.

src/manager.rs
ncloudioj: I'd recommend to do them both, i.e. having a workaround in rkv for the time being so that it works on Windows at least to get CI rolling. On the other hand, pop this back to upstream(s) in hope that this would get fixed in the right place. Yep, let me try it from lmdb-rs first, cuz I suspect that the windows mmap syscall won't play nice with UNC path, since it returned a error_no(123), which means it received an invalid directory according to its document.

Heh, I used to be on the lmdb's mailing list a while ago, and filed various bugs there, will try to grok their new issue tracking system if necessary :)

src/manager.rs
ncloudioj: Sure, will check it out.

src/manager.rs
ncloudioj: Haha, so you've also been there before. Such a small world! :)

PR #40, opened 2018-06-19
enable rustfmt

ncloudioj:
This replaces PR #32 in order to avoid those excessive merge conflicts. Also this adds various rustfmt configuration items based on the code review. Lastly, rustfmt check is enabled in TravisCi.

r? @mykmelez Could you take another look? You may notice there are a few long lines in this PR, since we've relaxed the `max_length` up to 120 characters, rustfmt decided to join some short lines to a a one-liner :). We can further tweak it around until we feel they're reasonable enough to get started.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -8,9 +8,7 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-use std::collections::{
-    BTreeMap,
-};
+use std::collections::BTreeMap;
 
 use std::io::{
     self,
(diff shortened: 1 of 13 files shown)


src/manager.rs
```diff
@@ -83,37 +73,48 @@ impl Manager {
...
+    pub fn get_or_create_with_capacity<'p, F, P>(
+        &mut self,
+        path: P,
+        capacity: c_uint,
+        f: F,
+    ) -> Result<Arc<RwLock<Rkv>>, StoreError>
+    where
+        F: FnOnce(&Path, c_uint) -> Result<Rkv, StoreError>,
+        P: Into<&'p Path>,
+    {
```
mykmelez: It looks like this is the only line that exceeds the new length limit, and it's indeed a really long line, so it's worth wrapping it.

ncloudioj:
@mykmelez Agreed, rustfmt is not yet perfect, we might want to overwrite its default behavior in some cases. Looks like this is an attribute `#[cfg_attr(rustfmt, rustfmt_skip)]` to let rustfmt skip the format, I guess we can use it to fix those quirks on-demand.

PR #41, opened 2018-06-19
run examples in travis

### LINKED ISSUE #36: run examples in automation
We should run the example programs in examples/ in automation to ensure we catch bustage.

ncloudioj:
This closes #36 (on Travis only) ;)

r? @mykmelez

### DIFF
--- a/run-all-examples.sh
+++ b/run-all-examples.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+cargo build --examples
+
+for file in examples/*; do
+    filename=$(basename ${file})
+    extension=${filename##*.}
+    example_name=${filename%.*}
+    if [[ "${extension}" = "rs" ]]; then
+        cargo run --example ${example_name}
+    fi
+done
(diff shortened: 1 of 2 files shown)


mykmelez:
I did some research on this for #42, and it seems like we can run all examples with the `--examples` flag on `cargo test`:

```
cargo test --examples
```

Alternately, the `--tests` flag seems to run both examples and unit/integration tests:

```
cargo test --tests
```

mykmelez:
> I did some research on this for #42, and it seems like we can run all examples with the --examples flag on cargo test:

Erm, I take that back. Doing so will run any tests in the examples, but it doesn't run the examples themselves!

run-all-examples.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+cargo build --examples
```
mykmelez: This is fine, although it shouldn't be necessary, as `cargo run --example` will build the example first.

mykmelez:
I'm going to merge this, but let's leave #36 open for a solution that works on Windows as well.

PR #58, opened 2018-07-20
Add support for multi-store read/write

### LINKED ISSUE #46: determine how to access multiple stores within a single transaction
Over in #42, I noted that "it's unclear if/how it's possible to open multiple stores within a single transaction, which LMDB itself supports," and @ncloudioj responded:

> 
> ```
> let txn = rkv.write();
> let store_foo = rkv.create_or_open("foo");
> let store_bar = rkv.create_or_open("bar");
> store_foo.write(txn, "key0", "value0");
> store_bar.write(txn, "key1", "value1");
> txn.commit();
> ```
> 
> The downside is that users can't use the Reader/Writer any more. Another potential approach, which reuses the design of Writer/Reader, is introduce a MultiStore so that multiple stores could be get_or_created at the same time in a single transaction. Its read/write API will be slightly different,
> 
> ```
> let store_names = vec!["foo", "bar", "baz"];
> let mega_store = rkv.create_or_open(&store_names);
> let writer = mega_store.write();
> writer.write("foo", "key0", "value0"); // it takes a store name here
> writer.write("bar", "key1", "value1");
> writer.commit();
> ```

This is a tough problem. The latter approach feels a bit more intuitive and is also likely to be more compact, provided store names are short; whereas the former grows a line for each store involved in the transaction.

On the other hand, the former approach has the advantage of being more strongly typed, because stores are referenced by handle after creation, so it isn't possible to compile code that opens the "foo" and "bar" stores and then writes to the "baz" store; whereas the latter approach will happily compile that code (and then fail at runtime).

Also note #29, although https://github.com/mozilla/rkv/pull/28#issuecomment-395448738 suggests that I didn't actually understand LMDB database handles when I filed it, and it's the wrong thing to do.

I'm also puzzling over the requirement that LMDB database handles be opened with reference to a specific transaction but can then be reused by any other transaction, as described in the docs for [mdb_dbi_open](http://www.lmdb.tech/doc/group__mdb.html#gac08cad5b096925642ca359a6d6f0562a), which additionally notes:

>The database handle will be private to the current transaction until the transaction is successfully committed. If the transaction is aborted the handle will be closed automatically. After a successful commit the handle will reside in the shared environment, and may be used by other transactions.
>
>This function must not be called from multiple concurrent transactions in the same process. A transaction that uses this function must finish (either commit or abort) before any other transaction in the process may use this function.

However, lmdb-rs appears to manage those constraints by acquiring a mutex and creating/committing a throwaway transaction in [Environment::create_db](https://github.com/danburkert/lmdb-rs/blob/90e88809954459869953cfa595ecd1e644bbfc87/src/environment.rs#L81-L105), so that shouldn't be an issue.

Regardless, from browsing the LMDB docs, it seems like the intent is for handles to stores to be long-lived, so perhaps the former approach is better, even though it doesn't let you use Reader/Writer, as it requires you to explicitly create the handles, which also enables you to reuse them.

Or perhaps even better is a related approach in which `Rkv::write` returns a non-store-specific `Writer` rather than an `lmdb::RwTransaction` (ditto for `Rkv::read`; it returns a `Reader`), and it has a `put` method that takes a store handle rather than a store name, i.e. something like:

```rust
let writer = rkv.write();
writer.put(store_foo, "key0", "value0");
writer.put(store_bar, "key1", "value1");
writer.commit();
// store_foo and store_bar can be reused to read
let reader = rkv.read();
reader.get(store_foo, "key0");
reader.get(store_bar, "key1");
```

(This has the added advantage that we no longer return the low-level RoTransaction and RwTransaction lmdb-rs structs from rkv methods.)

@ncloudioj What do you think?

ncloudioj:
This closes #46.

The implementation was based on [this](https://github.com/mozilla/rkv/issues/46#issuecomment-406422291)

r? @mykmelez

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -561,4 +561,163 @@ mod tests {
         //
         reader.iter_from("nuu").unwrap();
     }
+
+    #[test]
+    fn test_multiple_store_read_write() {
+        let root = Builder::new().prefix("test_multiple_store_reader_writer").tempdir().expect("tempdir");
+        fs::create_dir_all(root.path()).expect("dir created");
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        let main: Store<&str> = k.open_or_create("main").expect("opened");
+        let another: Store<&str> = k.open_or_create("another").expect("opened");
+        let third: Store<&str> = k.open_or_create("third").expect("opened");
+
+        // test put & put_in across multiple stores
+        let mut writer = main.write(&k).expect("writer");
+        writer.put("key", &Value::Str("value 0")).expect("wrote");
+        writer.put_in("key", &Value::Str("value 1"), &another).expect("wrote");
+        writer.put_in("key", &Value::Str("value 2"), &third).expect("wrote");
+
+        // test get & get_in on writer
+        assert_eq!(writer.get("key").expect("value"), Some(Value::Str("value 0")));
+        assert_eq!(writer.get_in("key", &main).expect("value"), Some(Value::Str("value 0")));
+        assert_eq!(writer.get_in("key", &another).expect("value"), Some(Value::Str("value 1")));
+        assert_eq!(writer.get_in("key", &third).expect("value"), Some(Value::Str("value 2")));
+        writer.commit().expect("committed");
+
+        // test get & get_in on reader
+        {
+            let reader = main.read(&k).expect("reader");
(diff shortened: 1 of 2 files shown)


ncloudioj:
Dup of #62

PR #60, opened 2018-07-21
Fixed the iterator example by using a bare vec to iterate

badboy:


mykmelez:
Hi @badboy, thanks for the fix! I'm curious what issue you encountered that was resolved by this change. Can you elaborate on the benefit of it?

badboy:
I'm pretty sure it didn't compile on my other machine, but now it does compile (current master, rustc 1.27.2), so I guess this change is not needed anymore (though seeing `&vec![...]` is always a bit weird :D)

mykmelez:
Hmm, I tried compiling on Rust versions back to 1.20.0 (which failed for unrelated reasons), and I couldn't reproduce the failure. Nevertheless, I agree that this version looks better. I'll merge it!

Issue #59, opened 2018-07-20
Deadlock: trying to open a store while a writer is in progress


ncloudioj:
Ran into this when I was testing #58, following snippet can reproduce the deadlock.

```rs
let store = rkv.open_or_create("store");
let writer = store.write();
let another_store = rkv.open_or_create("another_store");
```

The deadlock was caused by the fact that LMDB doesn't allow two write transactions created at the same time. In this case, the `writer` is a wrapper of `RwTransaction`. Under the hood, `open_or_create` will create another write transaction to open a store, it'll hang forever since `writer` will never get a chance to abort or commit. So the solution would be either open all the stores before spawning any writer, or commit the writer and then open other stores.

We should at least document this for `Store` to avoid this kind of deadlock.

mykmelez:
Nice catch!

We do currently document the potential for a call to _Store::write()_ to block on an extant Writer (in the [Basic Usage example](https://docs.rs/rkv/0.3.2/rkv/#basic-usage) and the [Store::write() API reference](https://docs.rs/rkv/0.3.2/rkv/struct.Store.html#method.write)). However, we don't currently document that _Rkv::open_or_create()_ will also block on an extant Writer, given the details of the underlying implementation. And that'll trip up API consumers. So we should definitely document it (and perhaps research ways to fix or work around the limitation, if it's feasible to do so).

ncloudioj:
This also reminds me that we should also allow a store to be opened in the read_only mode. Doing so allows the consumer to open the store to read while there is an in-flight write transaction on the same store.

This might require some changes in lmdb-rs.

mykmelez:
> This might require some changes in lmdb-rs.

Indeed. I've not had luck engaging with the lmdb-rs project recently. I have several PRs pending, the first two of which are now two months old. And you've received no response to your issue https://github.com/danburkert/lmdb-rs/issues/33. So I think we're going to need to fork the project and maintain our own fork, at least for the time being, in order to make this and other changes to it.

ncloudioj:
Agreed, let's fork it for now. Its code looks pretty clean and well-organized, should be easy for us to maintain and develop.

mykmelez:
> Agreed, let's fork it for now. Its code looks pretty neat and well-organized, should be easy for us to maintain and develop.

Ok, I've forked it to https://github.com/mozilla/lmdb-rs/ and added you as a collaborator!

PR #61, opened 2018-07-24
add an api to open a store with read transaction

ncloudioj:
This implements an API that opens a rkv store by a read transaction to avoid the deadlocks, see #59 for more details.

r? @mykmelez

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -123,6 +123,22 @@ impl Rkv {
         })?;
         Ok(Store::new(db))
     }
+
+    /// Open an existing database, unlike other `open_or_create_*` functions, it
+    /// opens the given database by using a read transaction, which means other
+    /// in-flight write transaction will not block this call. This is preferred
+    /// to be used in the read_only scenarios.
+    pub fn open<'s, T, K>(&self, name: T) -> Result<Store<K>, StoreError>
+    where
+        T: Into<Option<&'s str>>,
+        K: AsRef<[u8]>,
+    {
+        let db = self.env.open_db(name.into()).map_err(|e| match e {
+            lmdb::Error::BadRslot => StoreError::open_during_transaction(),
+            _ => e.into(),
+        })?;
+        Ok(Store::new(db))
+    }
 }
 
 /// Read and write accessors.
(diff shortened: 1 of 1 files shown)


mykmelez:
This looks good. Am I correct in thinking that it isn't possible to make Rkv::open_or_create() use a read transaction upstream, because database creation has to happen via a write transaction?

src/env.rs
```diff
@@ -123,6 +123,22 @@ impl Rkv {
...
+    /// opens the given database by using a read transaction, which means other
+    /// in-flight write transaction will not block this call. This is preferred
+    /// to be used in the read_only scenarios.
+    pub fn open<'s, T, K>(&self, name: T) -> Result<Store<K>, StoreError>
+    where
+        T: Into<Option<&'s str>>,
+        K: AsRef<[u8]>,
+    {
+        let db = self.env.open_db(name.into()).map_err(|e| match e {
+            lmdb::Error::BadRslot => StoreError::open_during_transaction(),
```
mykmelez: This isn't specific to your change, as it applies to the _Rkv::existing open_or_create*()_ methods, but I'm puzzled by the special handling of this error.

The lmdb-rs crate says of _Environment::open_db()_:

```
    /// This function will fail with `Error::BadRslot` if called by a thread which has an ongoing
    /// transaction.
```

And of _Environment::create_db()_:

```
    /// This function will fail with `Error::BadRslot` if called by a thread with an open
    /// transaction.
```

Which rkv translates into _StoreError::OpenAttemptedDuringTransaction_ ("attempted to open DB during transaction in thread").

But that seems to be what your test does, and it succeeds. Nor does _Rkv::existing open_or_create*()_ fail in that case; instead, it blocks. And we have no tests that trigger this error. So I wonder when we'll actually see it.

ncloudioj: This special error handling also puzzled me a lot. I think it has to do with [this](https://github.com/mozilla/rkv/blob/master/src/error.rs#L81). By default, `rkv::error` interprets it as `ReadTransactionAlreadyExists`, in this case, it wants to overwrite that with another interpretation.

I believe the `Error::BadRslot` is actually triggered by the transaction that being used to open the store. I managed to produce this error in this test:

```rs
 let _sk: Store<&str> = k.open_or_create("sk").expect("opened");
 let sk_readonly: Store<&str> = k.open("sk").expect("opened");
 let reader = sk_readonly.read(&k).expect("reader"); // this read transaction makes the following open panic with BadRslot
 let _sk_readonly_1: Store<&str> = k.open("sk").expect("opened");
```

Will add a test to cover this in a followup batch.

PR #62, opened 2018-07-26
Refactor to support multi-store

### LINKED ISSUE #46: determine how to access multiple stores within a single transaction

ncloudioj:
This is another try to fix #46 based on the discussion in #46.

This patch consists of following parts in 3 commits:
1. move `Reader&Writer` from `Store` to `Rkv`, also remove `get` function from `Store`
2. update all the examples
3. apply similar refactoring to integer store, note that this module appears to be a WIP, it needs some followup patches to support functionalities such as iterators

r? @mykmelez

### DIFF
--- a/src/integer.rs
+++ b/src/integer.rs
@@ -8,174 +8,153 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-// use std::marker::PhantomData;
+use std::marker::PhantomData;
 
-// use bincode::serialize;
+use bincode::serialize;
 
-// use lmdb::{
-// Database,
-// RoTransaction,
-// };
+use serde::Serialize;
 
-// use serde::Serialize;
+use error::{
+    DataError,
+    StoreError,
+};
 
-// use error::{
-// DataError,
-// StoreError,
-// };
+use value::Value;
 
-// use value::Value;
+use readwrite::{
+    Reader,
+    Store,
+    Writer,
+};
 
-// use readwrite::{
-// Reader,
-// Store,
-// Writer,
-// };
+pub trait EncodableKey {
+    fn to_bytes(&self) -> Result<Vec<u8>, DataError>;
+}
 
-// use Rkv;
+pub trait PrimitiveInt: EncodableKey {}
 
-// pub trait EncodableKey {
-// fn to_bytes(&self) -> Result<Vec<u8>, DataError>;
-// }
+impl PrimitiveInt for u32 {}
 
-// pub trait PrimitiveInt: EncodableKey {}
+impl<T> EncodableKey for T
+where
+    T: Serialize,
+{
+    fn to_bytes(&self) -> Result<Vec<u8>, DataError> {
+        serialize(self)         // TODO: limited key length.
+            .map_err(|e| e.into())
+    }
+}
 
-// impl PrimitiveInt for u32 {}
+pub(crate) struct Key<K> {
+    bytes: Vec<u8>,
+    phantom: PhantomData<K>,
+}
 
-// impl<T> EncodableKey for T
-// where
-// T: Serialize,
-// {
-// fn to_bytes(&self) -> Result<Vec<u8>, DataError> {
-// serialize(self)         // TODO: limited key length.
-// .map_err(|e| e.into())
-// }
-// }
+impl<K> AsRef<[u8]> for Key<K>
+where
+    K: EncodableKey,
+{
+    fn as_ref(&self) -> &[u8] {
+        self.bytes.as_ref()
+    }
+}
 
-// struct Key<K> {
-// bytes: Vec<u8>,
-// phantom: PhantomData<K>,
-// }
+impl<K> Key<K>
+where
+    K: EncodableKey,
+{
+    fn new(k: K) -> Result<Key<K>, DataError> {
+        Ok(Key {
+            bytes: k.to_bytes()?,
+            phantom: PhantomData,
+        })
+    }
+}
 
-// impl<K> AsRef<[u8]> for Key<K>
-// where
-// K: EncodableKey,
-// {
-// fn as_ref(&self) -> &[u8] {
-// self.bytes.as_ref()
-// }
-// }
+pub struct IntegerReader<'env, K>
+where
+    K: PrimitiveInt,
+{
+    inner: Reader<'env, Key<K>>,
+}
 
-// impl<K> Key<K>
-// where
-// K: EncodableKey,
-// {
-// fn new(k: K) -> Result<Key<K>, DataError> {
-// Ok(Key {
-// bytes: k.to_bytes()?,
-// phantom: PhantomData,
-// })
-// }
-// }
+impl<'env, K> IntegerReader<'env, K>
+where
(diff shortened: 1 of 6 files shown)


src/integer.rs
```diff
@@ -116,44 +108,26 @@ impl<'env, K> IntegerWriter<'env, K>
...
-    fn abort(self) {
-        self.inner.abort();
+    pub(crate) fn new(writer: Writer<Key<K>>) -> IntegerWriter<K> {
+        IntegerWriter {
+            inner: writer,
+        }
     }
-}
 
-impl<K> IntegerStore<K>
```
mykmelez: _Rkv::create_or_open_integer()_ still needs to return this type, and IntegerWriter/IntegerReader needs to require it. Otherwise, it'll be possible to retrieve a store with _Rkv::create_or_open_integer()_ and then write to it with the non-integer Writer:

```rust
    #[test]
    fn test_integer_store_with_non_integer_keys() {
        let root = Builder::new().prefix("test_integer_store_with_non_integer_keys").tempdir().expect("tempdir");
        fs::create_dir_all(root.path()).expect("dir created");
        let k = Rkv::new(root.path()).expect("new succeeded");
        let s = k.open_or_create_integer("s").expect("open");

        let mut writer = k.write().expect("writer");

        writer.put(&s, "123", &Value::Str("hello!")).expect("write");
        assert_eq!(writer.get(&s, "123").expect("read"), Some(Value::Str("hello!")));
    }
```

Which compiles and runs but doesn't retrieve the written value:

```
> > cargo test
> …
> test integer::tests::test_integer_store_with_non_integer_keys ... FAILED
> …
> failures:
> 
> ---- integer::tests::test_integer_store_with_non_integer_keys stdout ----
> 	thread 'integer::tests::test_integer_store_with_non_integer_keys' panicked at 'assertion failed: `(left == right)`
>   left: `None`,

```

Conversely, it'll be possible to retrieve a store with _Rkv::create_or_open()_ and then write to it with IntegerWriter, which compiles, runs, and retrieves the written value but probably isn't doing what the developer expects (i.e. applying LMDB's performance optimizations for integer-key stores).

The IntegerStore type enforces at compile time that all instances of IntegerWriter/IntegerReader are operating on a store that the developer retrieved via _Rkv::create_or_open_integer()_. I guess that still isn't entirely foolproof, as it's presumably still possible for the store in question to be a non-integer store (if it was previously created as such). But it's the equivalent of the current implementation.

(Ideally, I'd like Writer/Reader to be able to operate on both integer and non-integer stores, enforcing both of their key type constraints at compile time. But I'm not quite sure how to implement that. And in any case that's a separable problem. As you noted, IntegerStore is a WIP!)

ncloudioj: Nice catch, will take another look at this.

> Ideally, I'd like Writer/Reader to be able to operate on both integer and non-integer stores, enforcing both of their key type constraints at compile time.

I was also thinking about that possibility in the process of this code refactoring. Having one reader/writer to take care of both types would be great, but my previous experience of dealing with LMDB integer store in Python and Go told me it'd be hard to implement. Now let's see if we can figure this out in Rust.

ncloudioj:
Hmm, all the CI builds failed, and the failure doesn't make any sense to me.

Let me rebase against master, and try it again.

ncloudioj:
OK, it passed the CI build after the rebase. I think we can revert the change in #51 now. 

@mykmelez could you take another look at it?

mykmelez:
> OK, it passed the CI build after the rebase. I think we can revert the change in #51 now.

I updated the dependency to the 2018-07-18 nightly build in https://github.com/mozilla/rkv/commit/8c2cc3793664eafdea88fce5ef7c20ac427f15e7. I've been thinking that we'd continue to pin rustfmt to specific versions to avoid getting busted by future breaking changes. Although it might end up being more work to update periodically to use new features. What do you think?

> @mykmelez could you take another look at it?

Yup, looks great, thanks for the fixes!

PR #63, opened 2018-07-26
add a test to cover the bad_rslot error

ncloudioj:
A followup of #61 

r? @mykmelez

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -337,6 +337,23 @@ mod tests {
     }
 
     #[test]
+    fn test_open_fail_with_badrslot() {
+        let root = Builder::new().prefix("test_open_fail_with_badrslot").tempdir().expect("tempdir");
+        fs::create_dir_all(root.path()).expect("dir created");
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        // First create the store
+        let sk: Store<&str> = k.open_or_create("sk").expect("opened");
+        // Open a reader on this store
+        let _reader = sk.read(&k).expect("reader");
+        // Open the same store for read while the reader is in progress will panic
+        let store: Result<Store<&str>, StoreError> = k.open("sk");
+        match store {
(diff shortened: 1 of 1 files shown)


src/env.rs
```diff
@@ -336,6 +336,25 @@ mod tests {
...
 
+    #[test]
+    fn test_open_fail_with_badrslot() {
+        let root = Builder::new().prefix("test_open_fail_with_badrslot").tempdir().expect("tempdir");
+        fs::create_dir_all(root.path()).expect("dir created");
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        // First create the store
+        let _sk: Store<&str> = k.open_or_create("sk").expect("opened");
+        // Open the same store for read
+        let sk_readonly: Store<&str> = k.open("sk").expect("opened");
```
mykmelez: Ah, interesting. It looks like you don't even need to open it this second time. Skipping this step still reproduces the problem for me.

mykmelez: I.e.:

```rust
        // First create the store
        let sk: Store<&str> = k.open_or_create("sk").expect("opened");
        // Open a reader on this store
        let _reader = sk.read(&k).expect("reader");
        // Open the same store for read while the reader is in progress will panic
        let store: Result<Store<&str>, StoreError> = k.open("sk");
```

@ncloudioj Can you reproduce the error with these steps?

src/env.rs
```diff
@@ -336,6 +336,25 @@ mod tests {
...
+        // Open a reader on this store
+        let _reader = sk_readonly.read(&k).expect("reader");
+        // Open the same store for read while the reader is in progress will panic
```
mykmelez: Technically it'll return an error rather than panicking, right?

src/env.rs
ncloudioj: @mykmelez You're completely right, as long as there is an ongoing reader, the underlying read transaction in `Rkv::open()` would hit "BAD_SLOT".

Already updated the patch.

PR #66, opened 2018-07-31
derive Copy for Store (fix #29)

### LINKED ISSUE #29: grow Manager to hand out Arc<Store> and ensure Store itself is Sync
Per the conversation in #28, the Manager should grow to conserve and hand out `Arc<Store>` instances and ensure that the `Store` itself is `Sync` (or otherwise help the consumer to satisfy the constraint from  http://www.lmdb.tech/doc/ to "not have open an LMDB database twice in the same process at the same time.").

mykmelez:
@ncalexan Per @rnewman's comment https://github.com/mozilla/rkv/pull/28#issuecomment-395448738, Store can be Copy, and access to it doesn't need to be synchronized, as described by #29. So this branch derives Copy for Store and tests access to it by multiple threads.

### DIFF
--- a/src/readwrite.rs
+++ b/src/readwrite.rs
@@ -163,6 +163,7 @@ impl<'env> Iterator for Iter<'env> {
 }
 
 /// Wrapper around an `lmdb::Database`.
+#[derive(Copy, Clone)]
 pub struct Store {
     db: Database,
 }
--- a/src/readwrite.rs
+++ b/src/readwrite.rs
@@ -162,7 +162,8 @@ impl<'env> Iterator for Iter<'env> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/readwrite.rs
```diff
@@ -163,6 +163,7 @@ impl<'env> Iterator for Iter<'env> {
 }
 
 /// Wrapper around an `lmdb::Database`.
+#[derive(Copy, Clone)]
```
ncalexan: I think it would be worth commenting that at this time, the underlying LMDB handle is a C integer, and therefore `Copy` is automatic :)

mykmelez: @ncalexan Noted in 5ae8508!

PR #77, opened 2018-08-15
Update uuid to 0.6 and bump version

Eijebong:


mykmelez:
Hi @Eijebong, thanks for the PR! Is there something in particular that we get with version 0.6? I ask because Gecko is using 0.5, and there's some value in aligning dependencies between the two, given that Gecko is rkv's largest known consumers. So I'll want to update Gecko (the next time we import rkv into mozilla-central) if we update rkv.

Eijebong:
Hey, it's to update it in gecko to start the work on deduping rand ;)

mykmelez:
Aha, in that case, sounds great! I'll publish a new version of rkv to crates.io after landing this so you can vendor the change into mozilla-central.

One nit: it looks like this isn't a breaking change to the API for rkv, so we should update rkv's "patch" version part rather than the "minor" version part, i.e. update it to 0.4.2 rather than 0.5.0.

Eijebong:
It's exposed in the public API

lib.rs:
`pub use value::Value;`

And
value.rs
`use uuid::UuidBytes;`
`pub enum Value<'s> { ..., Uuid(&'s UuidBytes), ... }`

So unless I'm missing something here, this is unfortunately a breaking change.

mykmelez:
> So unless I'm missing something here, this is unfortunately a breaking change.

Ah, I was the one who was missing something! Indeed, you are correct. Ok, then this looks good, so I'll land and publish it.

mykmelez:
Published! https://crates.io/crates/rkv/0.5.0

mykmelez:
@Eijebong Any chance you have a link to the Bugzilla bug or Phabricator patch where you're updating uuid in Gecko?

I've landed #79 to dedupe new-ordered-float/ordered-float, which requires vendoring rkv into mozilla-central. But doing so now would dupe uuid for versions 0.5 and 0.6. So I'm wondering about your timeline for updating Gecko to use uuid 0.6 consistently.

Eijebong:
I totally forgot about that, sorry. Been way too busy updating hyper :/
I'll try to do it this week end if you didn't get to it before

mykmelez:
I had some time today to update both rkv and uuid in Gecko, so I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1500259 on updating rkv and submitted https://phabricator.services.mozilla.com/D9160 to update rkv to 0.5 and uuid to 0.6.

(Just in time for uuid 0.7 to be released a few days ago. 😉 )

PR #79, opened 2018-09-11
Switch from new-ordered-float to ordered-float

mbrubeck:
The `new-ordered-float` fork has been merged back into the original `ordered-float` crate.


mykmelez:
@mbrubeck Thanks for the fix! Is there an equivalent for the servo/components/style/ directory in mozilla-central? I'd like to keep rkv dependencies aligned with those of Gecko as much as possible, to avoid redundancy when vendoring rkv into mozilla-central.

mbrubeck:
See servo/servo#21681.  @emilio has been merging changes from the servo repo to mozilla-central periodically.  (This is very tiny dependency so it's probably fine if it gets temporarily out of sync, though.)

emilio:
Yeah, I sync back roughly every week, more or less.

emilio:
I just did the sync in https://bugzilla.mozilla.org/show_bug.cgi?id=1491577.

mykmelez:
I published the new version of rkv on September 19, but it took longer than expected to update Gecko to that version, as I've been waiting for @Eijebong to update uuid in Gecko, as we discussed in #77.

I had some time today to do both, however, so I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1500259 on updating rkv and submitted https://phabricator.services.mozilla.com/D9160 to update rkv to 0.5 and uuid to 0.6.

(Just in time for uuid 0.7 to be released a few days ago. 😉 )

Issue #34, opened 2018-06-07
Adding mechanisms to gracefully handle "map is full" error


ncloudioj:
Each lmdb store has a predefined size (10MB by default), if a store is running out of free space due to,
* Store was filled by the data
* Some orphan transactions prevented lmdb from reclaiming the unused pages 

either way, all the following inserts will be rejected by lmdb with a `MDB_MAP_FULL` error.

We will have to provider some bailout mechanisms for this particular issue to avoid the write downtime.
* Resizing the store, this requires all users terminate their transactions (read/write) first, also, once the size gets increased, there is no way to shrink it unless recreating a new one and copying all the data over
* To let lmdb reclaim the unused pages, we need to ensure there is no orphan holding locks in lmdb's transaction table. Lmdb has a API (mdb_reader_check) to clean up those zombie transactions, looks like this API was not exposed by ldmb-rs, we might have to add that to upstream first.

shivawu:
On a related matter, it seems that `mdb_env_set_mapsize` should be exposed to change the default size. I went back to lower level library because of this limitation. Not sure if there's an existing issue on that.

mykmelez:
@shivawu You can change the map size using rkv. See #82 for tests that demonstrate how to do that.

shivawu:
Ah I see, somehow I missed it. Thanks for pointing that out. Probably the local doc does not work that well for me.. Also btw, the online doc is missing for the project (https://docs.rs/rkv/0.5.1/rkv/), expected?

mykmelez:
> Ah I see, somehow I missed it. Thanks for pointing that out. Probably the local doc does not work that well for me.. Also btw, the online doc is missing for the project (https://docs.rs/rkv/0.5.1/rkv/), expected?

Not expected, but known. It's #81, where you can read all the details (tl;dr: docs.rs is using an outdated rustc compiler).

ncloudioj:
Another way for us to deal with MAP_FULL is that we check the space usage of environment via its stats API (which reports max map size, total page # in use, total page # in the freelist etc), and then increase the map size if the disk usage goes over a pre-defined high watermark.

Specifically, we can do this either by a dedicated maintenance worker, or by the consumers themselves. Either way, some coordination is required since resizing can only be done on a free environment (only one writer without any readers).

This preventive resizing appears to be more preferred over the resize-when-you-hit-the-map-full in two aspects:

* Write transactions are less likely to be interrupted
[...]

PR #82, opened 2018-10-19
test exceeding and increasing the map size

mykmelez:
Per the comment from @shivawu in #34 that there should be a way to set an environment's map size, here are tests that demonstrate how to do that.

NB: We should file upstream bugs/PRs in the LMDB crate and C library to fix the code or docs (probably the latter) to agree about the default map size for an LMDB environment.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -278,6 +278,61 @@ mod tests {
     }
 
     #[test]
+    #[should_panic(expected = "wrote: LmdbError(MapFull)")]
+    fn test_exceed_map_size() {
+        let root = Builder::new().prefix("test_exceed_map_size").tempdir().expect("tempdir");
+        println!("Root path: {:?}", root.path());
+        fs::create_dir_all(root.path()).expect("dir created");
+        assert!(root.path().is_dir());
+
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        let sk: Store = k.open_or_create_default().expect("opened");
+
+        // Writing a large enough value should cause LMDB to fail on MapFull.
+        //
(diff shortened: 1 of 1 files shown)


mykmelez:
@rnewman How does this look?

(Perhaps it makes sense to add an `Rkv::with_size()` constructor function that sets the map size, just as we have `Rkv::with_capacity()` for setting the database capacity.)

src/env.rs
```diff
@@ -277,6 +277,61 @@ mod tests {
...
+        let root = Builder::new().prefix("test_open_with_map_size").tempdir().expect("tempdir");
+        println!("Root path: {:?}", root.path());
+        fs::create_dir_all(root.path()).expect("dir created");
+        assert!(root.path().is_dir());
+
+        let mut builder = Rkv::environment_builder();
+        builder.set_map_size(1024 * 1024 * 10 /* 10,485,760 bytes, i.e. 10MiB */);
+        let k = Rkv::from_env(root.path(), builder).unwrap();
+        let sk: Store = k.open_or_create_default().expect("opened");
+        let val = "x".repeat(1024 * 1024 /* 1,048,576 bytes, i.e. 1MiB */);
```
rnewman: You should make sure that this test and the one above use the same value or size (maybe `fn get_larger_than_default_map_size_value()`), and are thus related more closely. Consider a year from now when the default bumps to 2MB, and/or `set_map_size` accrues a bug. The first test will break, and this one will not, because this value is now smaller than the limit.

mykmelez: Good point, I've refactored references to the default map size into a get_larger_than_default_map_size_value() function that both tests use to determine a value slightly larger than the default map size.

PR #83, opened 2018-10-23
implement conversion from Value to OwnedValue and expose OwnedValue

mykmelez:
There's this comment on the OwnedValue implementation:

> // TODO: implement conversion between the two types of `Value` wrapper.
> // This might be unnecessary: we'll probably jump straight to primitives.

However, I've identified a use case: in Gecko, structs that implement an XPCOM interface may want to store a `Value`, but they can't hold a reference, because the XPCOM Rust binding doesn't support that; so they would need to store an `OwnedValue` instead (or generalize over the primitive type and make their consumers convert the reference types to owned types).

This branch implements the `Value` to `OwnedValue` conversion for storage in XPCOM structs (and any other unknown unknown use cases out there).

It also upgrades the uuid dependency from 0.6 to 0.7, which isn't strictly necessary but allows us to make the conversion infallible, since the `Value::Uuid` to `OwnedValue::Uuid` conversion calls `Uuid::from_bytes()`, which is [fallible (at runtime)](https://docs.rs/uuid/0.6.5/uuid/struct.Uuid.html#method.from_bytes) in version 0.6 but [infallible (at runtime, fails at compile-time)](https://docs.rs/uuid/0.7.1/uuid/struct.Uuid.html#method.from_bytes) in version 0.7.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ bincode = "1.0"
 lazy_static = "1.0"
 lmdb-rkv = "0.8"
 ordered-float = "1.0"
-uuid = "0.6"
+uuid = "0.7"
 serde = "1.0"
 url = "1.7.0"
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,4 +218,7 @@ pub use readwrite::{
     Writer,
 };
 
-pub use value::Value;
+pub use value::{
+    OwnedValue,
+    Value,
+};
--- a/src/value.rs
+++ b/src/value.rs
@@ -180,10 +180,6 @@ impl<'s> Value<'s> {
             },
         }.map_err(DataError::EncodingError)
     }
-
-    pub fn to_owned(&self) -> OwnedValue {
-        self.into()
-    }
 }
 
 impl<'s> From<&'s Value<'s>> for OwnedValue {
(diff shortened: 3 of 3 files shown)


mykmelez:
> This looks fine to me. Is there an `impl From<Value> for OwnedValue` that should be declared?

There can be, and it seems more idiomatic. I've implemented that in 074ca00. Given that impl, however, it isn't clear if we still want `Value.to_owned()`, which now just returns `self.into()`. Shall I get rid of it and have consumers use `OwnedValue::from()` or `Value.into()` exclusively?

ncalexan:
> > This looks fine to me. Is there an `impl From<Value> for OwnedValue` that should be declared?
> 
> There can be, and it seems more idiomatic. I've implemented that in [074ca00](https://github.com/mozilla/rkv/commit/074ca0053b81516b227492130980c974a31db7c3). Given that impl, however, it isn't clear if we still want `Value.to_owned()`, which now just returns `self.into()`. Shall I get rid of it and have consumers use `OwnedValue::from()` or `Value.into()` exclusively?

I would.

ncalexan:
[074ca00](https://github.com/mozilla/rkv/commit/074ca0053b81516b227492130980c974a31db7c3) looks good.

mykmelez:
> I would.

Ok, I've done so in 2fceb32. I'll land this branch once it passes CI.

PR #93, opened 2018-11-29
RFC - Implementation of MultiReader and MultiWriter

rrichardson:
This is a proposal for the MultiReader and MultiWriter interface.  It is rather similar to the regular Reader and Writer except that it uses the  iter_dup_* family of cursor ops.  `get` returns an Iter instead of a `Value`. 

The biggest change is the introduction of what I am tentatively calling `MultiCursor` which is basically just a `MultiReader`, but built around the `RWTransaction`.  
The way you get a `MultiCursor` is call `MultiReader::into_cursor` which does what you'd expect.  and `MultiCursor::into_writer` will restore the writer.  This API design is intended to keep the user from intermixing reads and writes and causing problems with the Multiple values. 

I still don't have a good solution for lifetimes, it seems a bit awkward to use a writer within a function where ids/values are generated.  

https://github.com/mozilla/rkv/pull/93/files#diff-012b062f0cc023a61bfe77ff53148a2eR24

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ backtrace = ["failure/backtrace", "failure/std"]
 arrayref = "0.3"
 bincode = "1.0"
 lazy_static = "1.0"
-lmdb-rkv = { path = "../lmdb-rs" }
+lmdb-rkv = { git="https://github.com/rrichardson/lmdb-rs", branch="iter-dup-test" }
 ordered-float = "1.0"
 uuid = "0.7"
 serde = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ backtrace = ["failure/backtrace", "failure/std"]
 arrayref = "0.3"
 bincode = "1.0"
 lazy_static = "1.0"
-lmdb-rkv = "0.9"
+lmdb-rkv = { path = "../lmdb-rs" }
 ordered-float = "1.0"
 uuid = "0.7"
 serde = "1.0"
--- a/src/multirw.rs
+++ b/src/multirw.rs
@@ -12,16 +12,8 @@ use lmdb;
 
 use std::marker::PhantomData;
 use std::rc::Rc;
-use lmdb::{
-    Cursor,
-    Database,
-    Iter as LmdbIter,
-    IterDup as LmdbIterDup,
-    RoCursor,
-    RoTransaction,
-    RwTransaction,
-    Transaction,
-};
+use lmdb::{Cursor, Database, Iter as LmdbIter, IterDup as LmdbIterDup, RoCursor, RoTransaction, RwTransaction,
+           Transaction};
 
 use lmdb::WriteFlags;
 
@@ -31,7 +23,11 @@ use value::Value;
 
 fn read_transform(val: Result<&[u8], lmdb::Error>) -> Result<Option<Value>, StoreError> {
     match val {
-        Ok(bytes) => Value::from_tagged_slice(bytes).map(Some).map_err(StoreError::DataError),
+        Ok(bytes) => {
+            Value::from_tagged_slice(bytes).map(Some).map_err(
+                StoreError::DataError,
+            )
+        },
         Err(lmdb::Error::NotFound) => Ok(None),
         Err(e) => Err(StoreError::LmdbError(e)),
     }
@@ -62,9 +58,9 @@ pub struct Iter<'env> {
     iter: LmdbIter<'env>,
 }
 
-pub struct MultiCursor<'env, K> 
+pub struct MultiCursor<'env, K>
 where
-    K: AsRef<[u8]>
+    K: AsRef<[u8]>,
 {
     tx: RwTransaction<'env>,
     phantom: PhantomData<K>,
@@ -72,21 +68,25 @@ where
 
 impl<'env, K> MultiCursor<'env, K>
 where
-    K: AsRef<[u8]> 
+    K: AsRef<[u8]>,
 {
-
     /// Provides a cursor to all of the values for the duplicate entries that match this key
     pub fn get(&self, store: MultiStore, k: K) -> Result<Iter, StoreError> {
-        let mut cursor = self.tx.open_ro_cursor(store.0).map_err(StoreError::LmdbError)?;
+        let mut cursor = self.tx.open_ro_cursor(store.0).map_err(
+            StoreError::LmdbError,
+        )?;
         let iter = cursor.iter_dup_of(k);
         //Ok(Iter{ iter, cursor })
-        Ok(Iter{ iter })
+        Ok(Iter { iter })
     }
 
     /// Consume this MultiCursor and give the `MultiWriter` back
     /// So that it may perform additional tasks
     pub fn into_writer(self) -> MultiWriter<'env, K> {
-        MultiWriter { tx: self.tx, phantom: PhantomData }
+        MultiWriter {
+            tx: self.tx,
+            phantom: PhantomData,
+        }
     }
 }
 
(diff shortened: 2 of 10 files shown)


rrichardson:
Hmm.. I used rustfmt because the checks failed formatting and it went and reformatted everything..  
https://github.com/mozilla/rkv/pull/93/files#diff-28db2d20bd1b4aee545371ed55845811 is the interesting one.

src/multirw.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn into_cursor(self) -> MultiCursor<'env, K> {
+        MultiCursor {
+            tx: self.tx,
+            phantom: PhantomData,
+        }
+    }
+
+    /// Insert a value at the specified key.
+    /// This put will allow duplicate entries.  If you wish to have duplicate entries
+    /// rejected, use the `put_flags` function and specify NO_DUP_DATA
```
mykmelez: Nit: presumably `put_flags` -> `put_with_flags`.

src/multirw.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn get(&self, store: MultiStore, k: K) -> Result<Iter, StoreError> {
+        let mut cursor = self.tx.open_ro_cursor(store.0).map_err(
+            StoreError::LmdbError,
+        )?;
+        let iter = cursor.iter_dup_of(k);
+        //Ok(MultiIter { iter, cursor, })
+        Ok(Iter { iter })
+    }
+
+    /// Cancel this read transaction (not particularly useful)
```
mykmelez: The LMDB [Getting Started](http://www.lmdb.tech/doc/starting.html) page talks about the utility of aborting read transactions:

> For read-only transactions, obviously there is nothing to commit to storage. The transaction still must eventually be aborted to close any database handle(s) opened in it, or committed to keep the database handles around for reuse in new transactions.
>
> In addition, as long as a transaction is open, a consistent view of the database is kept alive, which requires storage. A read-only transaction that no longer requires this consistent view should be terminated (committed or aborted) when the view is no longer needed (but see below for an optimization).

src/multirw.rs
```diff
@@ -0,0 +1,224 @@
...
+{
+    pub(crate) fn new(txn: RwTransaction) -> MultiWriter<K> {
+        MultiWriter {
+            tx: txn,
+            phantom: PhantomData,
+        }
+    }
+
+    /// This cursor consumes the writer, as it is not safe to attempt to access records while using
+    /// put or delete, as the location of the records might change
```
mykmelez: It's possible to write records using a cursor in LMDB, although rkv doesn't yet support this. Perhaps it would be more ergonomic to support that functionality than to make the consumer convert back and forth between the MultiCursor and the MultiWriter. Then we could keep `MultiWriter.into_cursor()`, which consumes the MultiWriter, but expose `MultiCursor.put()/delete()` methods, so it isn't necessary to convert back to MultiWriter via `MultiCursor.into_writer()` in order to make changes.

mykmelez:
> Hmm.. I used rustfmt because the checks failed formatting and it went and reformatted everything..

Hmm, strange, we have this [.rustfmt.toml configuration file](https://github.com/mozilla/rkv/blob/master/.rustfmt.toml) and check the formatting of patches in (Travis) CI.

I did just notice that the latest nightly build of Rust suggests a few formatting changes that the pinned nightly build of Rust in CI doesn't catch. I've submitted a PR to fix those nits and switch back to the latest nightly build of Rust in CI in #94.

But that still doesn't explain why you're seeing so many changes. Is your invocation of rustfmt perhaps not honoring the .rustfmt.toml file in tree? Many of the formatting changes in this branch seem to be reducing line lengths by wrapping more aggressively, and we have a relatively lenient line-length limit, `max_width = 120`, plus `use_small_heuristics = "Off"` also affects line lengths in some cases.

Cargo.toml
```diff
@@ -18,7 +18,7 @@ backtrace = ["failure/backtrace", "failure/std"]
 arrayref = "0.3"
 bincode = "1.0"
 lazy_static = "1.0"
-lmdb-rkv = "0.9"
+lmdb-rkv = { path = "../lmdb-rs" }
```
mykmelez: Did you wind up needing to make any changes to _lmdb-rs_ in order to support MultiReader/Writer?

rrichardson: Yes.  A terribly minor change: 
In https://github.com/mozilla/lmdb-rs/blob/master/src/cursor.rs#L83  I changed the key param from `&K` to `K` .  The reference seemed redundant, and this way it matches the `iter` and `iter_from` functions.

src/multirw.rs
rrichardson: I agree, I still think we should test with the cursor to ensure that it's safe to get->put dup values in LMDB while ensuring that the retrieved value is still in-tact.  I did a significant amount of work with LMDB about 5 years ago, and I vaguely recall this exact problem, but maybe the API has been updated since then. 

Adding Cursor support would dovetail in very nicely with my other proposed API change: https://github.com/mozilla/rkv/issues/95

src/multirw.rs
rrichardson: rgr that.  I'll remove the contents in the parens :)

rrichardson:
I have a hunch that I ran fustfmt in a way that it didn't honor the `.rustfmt.toml` so it used more strict defaults.  Re-running rustfmt using the correct settings then doesn't do anything.

rrichardson:
OK.. I ran `cargo fmt` . after getting the latest rustfmt nightly.. It is quite a bit better (still not identical)

mykmelez:
> OK.. I ran `cargo fmt` . after getting the latest rustfmt nightly.. It is quite a bit better (still not identical)

You may have to run `cargo +nightly fmt` to get the latest rustfmt nightly formatting behavior. I did that and saw a few changes, which I've put into https://github.com/rrichardson/rkv/pull/1.

src/multirw.rs
```diff
@@ -0,0 +1,256 @@
...
+        // TODO: don't allocate twice.
+        let bytes = v.to_bytes()?;
+        self.tx.put(store.0, &k, &bytes, flags).map_err(StoreError::LmdbError)
+    }
+    
+    pub fn delete_all(&mut self, store: MultiStore, k: K) -> Result<(), StoreError> {
+        self.tx.del(store.0, &k, None).map_err(StoreError::LmdbError)
+    }
+
+    pub fn delete(&mut self, store: MultiStore, k: K, v: &Value) -> Result<(), StoreError> {
```
mykmelez: Note that there's an unimplemented `Writer.delete_value()` stub in writer.rs (that suggests it should be a method on a DUP_SORT store). Now that we're implementing this approach, we should remove that stub.

src/multirw.rs
```diff
@@ -0,0 +1,256 @@
...
+        let mut cursor = self.tx.open_ro_cursor(store.0).map_err(StoreError::LmdbError)?;
+        let iter = cursor.iter_dup_of(k);
+        //Ok(MultiIter { iter, cursor, })
+        Ok(Iter {
+            iter,
+            cursor,
+        })
+    }
+
+    /// Provides a cursor to all of the values for the duplicate entries that match this key
```
mykmelez: This comment looks like a dupe of the one above, would need updating to reflect the difference between the two functions.

src/multirw.rs
```diff
@@ -0,0 +1,256 @@
...
+{
+    /// Provides a cursor to all of the values for the duplicate entries that match this key
+    pub fn get(&self, store: MultiStore, k: K) -> Result<Iter, StoreError> {
+        let mut cursor = self.tx.open_ro_cursor(store.0).map_err(StoreError::LmdbError)?;
+        let iter = cursor.iter_dup_of(k);
+        Ok(Iter{ iter, cursor })
+    }
+
+    /// Provides a cursor to all of the values for the duplicate entries that match this key
+    pub fn get_first(&self, store: MultiStore, k: K) -> Result<Option<Value>, StoreError> {
```
mykmelez: This comment too is a duplicate of the one above and needs updating.

src/multirw.rs
```diff
@@ -0,0 +1,256 @@
...
+            None => None,
+            Some((key, bytes)) => Some((key, read_transform(Ok(bytes)))),
+        }
+    }
+}
+
+/// New type around an `lmdb::Database`.  At this time, the underlying LMDB
+/// handle (within lmdb-rs::Database) is a C integer, so Copy is automatic.
+#[derive(Copy, Clone)]
+pub struct MultiStore(Database);
```
mykmelez: Should we make the *Database* inside the *Store* struct be `pub(crate)` and have *MultiStore* wrap a *Store*, like *IntegerStore* does?

rrichardson:
I'd like to defer correcting this until we decide what the new API will look like in #95 

I'm not really happy with this API.. I would be happy with either option in #95.  

Note that nothing in #95 solves the LMDB Cursor issues.

rrichardson:
Closing this in favor of https://github.com/mozilla/rkv/pull/97 to nail down an API.

PR #99, opened 2018-12-17
test consumer-specified integer key variants

mykmelez:
@ncloudioj This just tests that we can generalize across integer types with generic type parameters. It contains only test changes, no functional changes. I created it to help think through the changes we're contemplating in #97, but it seems worth landing as well, to document (and verify) that this works.

Let me know what you think!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ lmdb-rkv = "0.9"
 ordered-float = "1.0"
 uuid = "0.7"
 serde = "1.0"
+serde_derive = "1.0"
 url = "1.7.0"
 
 # Get rid of failure's dependency on backtrace. Eventually
--- a/src/integer.rs
+++ b/src/integer.rs
@@ -160,18 +160,16 @@ mod tests {
         let s = k.open_or_create_integer("s").expect("open");
 
         macro_rules! test_integer_keys {
-            ($type:ty, $key:expr) => {
-                {
-                    let mut writer = k.write_int::<$type>().expect("writer");
+            ($type:ty, $key:expr) => {{
+                let mut writer = k.write_int::<$type>().expect("writer");
 
-                    writer.put(s, $key, &Value::Str("hello!")).expect("write");
-                    assert_eq!(writer.get(s, $key).expect("read"), Some(Value::Str("hello!")));
-                    writer.commit().expect("committed");
+                writer.put(s, $key, &Value::Str("hello!")).expect("write");
+                assert_eq!(writer.get(s, $key).expect("read"), Some(Value::Str("hello!")));
+                writer.commit().expect("committed");
 
-                    let reader = k.read_int::<$type>().expect("reader");
-                    assert_eq!(reader.get(s, $key).expect("read"), Some(Value::Str("hello!")));
-                }
-            }
+                let reader = k.read_int::<$type>().expect("reader");
+                assert_eq!(reader.get(s, $key).expect("read"), Some(Value::Str("hello!")));
+            }};
         }
 
         test_integer_keys!(u32, 123);
--- a/src/integer.rs
+++ b/src/integer.rs
@@ -159,13 +159,30 @@ mod tests {
         let k = Rkv::new(root.path()).expect("new succeeded");
         let s = k.open_or_create_integer("s").expect("open");
 
-        let mut writer = k.write_int::<u32>().expect("writer");
+        macro_rules! test_integer_keys {
+            ($type:ty, $key:expr) => {
+                {
+                    let mut writer = k.write_int::<$type>().expect("writer");
 
-        writer.put(s, 123, &Value::Str("hello!")).expect("write");
-        assert_eq!(writer.get(s, 123).expect("read"), Some(Value::Str("hello!")));
-        writer.commit().expect("committed");
+                    writer.put(s, $key, &Value::Str("hello!")).expect("write");
+                    assert_eq!(writer.get(s, $key).expect("read"), Some(Value::Str("hello!")));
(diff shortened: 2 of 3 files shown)


ncloudioj:
r+, nice abstraction!

A question - Other than u32, shall we also implement `PrimitiveInt` for other integer types?

mykmelez:
> A question - Other than u32, shall we also implement `PrimitiveInt` for other integer types?

Good question! Yes, I think we should, as it isn't as trivial as I thought to implement it for other integer types. Attempting to do so outside of the integer module the way I did it in the first commit on this branch—via a simple `impl` item—results in an error like:

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
  --> tests/integer-store.rs:65:5
   |
65 |     impl PrimitiveInt for u64 {}
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ impl doesn't use types inside crate
   |
   = note: the impl does not reference any types defined in this crate
   = note: define and implement a trait or new type instead
```

It's still possible to do so, if you define a newtype, implement `PrimitiveInt`, and implement or derive `serde::Serialize` for it. But then you have to use that newtype in place of the primitives.

I've added an integration test in tests/ for this, however I've also implemented `PrimitiveInt` internally for `u16` and `u64` in addition to `u32`, so consumers can use those key types without further ado.

(I omitted `u8`, which seems unlikely to be useful, although you never know. And I omitted `u128`, which panics at runtime with "write: DataError(EncodingError(Custom("u128 is not supported. Enable the `i128` feature of `bincode`")))".)

Finally, I made the `IntegerWriter.abort()` and `.commit()` methods `pub`, as I expect they were intended to be (since you can't commit a transaction without the latter).

mykmelez:
Having implemented support for `u16` and `u64` keys, I now see from http://www.lmdb.tech/doc/group__mdb__dbi__open.html#ga219c5c50d41902fca79c68d8836e7753 that LMDB defines MDB_INTEGERKEY as "numeric keys in native byte order: either unsigned int or size_t."

So perhaps it's incorrect to support the `u64` type, because it gets wrapped (or has other undefined behavior) when LMDB stores it. And perhaps it's misleading to support the `u16` type, because LMDB stores it as u32 (the Rust equivalent of C `int`) anyway.

@ncloudioj, what do you think?

ncloudioj:
> So perhaps it's incorrect to support the u64 type, because it gets wrapped (or has other undefined behavior) when LMDB stores it. And perhaps it's misleading to support the u16 type, because LMDB stores it as u32 (the Rust equivalent of C int) anyway.

I believe that LMDB is able to handle all the primitive integers from i16/u16 and beyond, provided that the user specifies the correct size, i.e. `mdb_value.mv_size`. Internally, it treats all the integers as two-byte aligned bytes in the native order. See more details [here](https://github.com/LMDB/lmdb/blob/mdb.master/libraries/liblmdb/mdb.c#L5697-L5745). As such, it'd only occupy exact two bytes on `u16` if `PrimitiveInt::<u16>::to_bytes` fulfills so.

Note that as an optimization, LMDB could fall back to the native unsigned integer (u32 and size_t) comparisons in certain cases, therefore, misuse of the signed and unsigned key could lead to some unexpected behaviors if the stored key and the lookup key share the same bit representation.

To minimize the confusion and complexity, to begin with, I'd recommend us to only expose `i32` & `u32` for now, with some notes on the proper integer and unsigned integer usage on the library consumers. For instance, it's up to them to make sure the type is consistent on read & write. If they create a store with `i32` integer keys, but later open it to read as `u32`, they might get some incorrect results (is it too picky? :).

We can further expose 2-byte and 8-byte integers in the future if necessary. Lastly, since `i8` and `u8` is rarely used in practice, and to implement it would require us to use the user-define comparators in LMDB, I think we can leave them out.

Does it make sense? @mykmelez

mykmelez:
> To minimize the confusion and complexity, I'd recommend us to only expose `i32` & `u32` to begin with, with some notes to the library consumers on the usage of integer and unsigned integer keys. For instance, it's up to them to make sure the type is consistent on read & write. If they create a store with `i32` integer keys, but later open it to read as `u32`, they might get some incorrect results (is it too picky? :).

It isn't too picky! In fact I think it isn't picky enough, given that rkv is intended to be a high-level interface that avoids footguns, and supporting both `i32` and `u32` keys means that a single-character typo could cause silent failures with no errors nor warnings (neither at compile time nor runtime).

After all, we explicitly say in the [rkv docs](https://docs.rs/rkv/) that rkv uses "Rust's type system to make single-typed key stores (including LMDB's own integer-keyed stores) safe and ergonomic." And supporting `i32` keys, while it may be ergonomic, doesn't seem particularly safe.

Thus I now think that we shouldn't implement support for other integer types besides `u32`. And consumers who want such support should define a newtype and implement `PrimitiveInt` for it (or use the low-level LMDB API directly).

> We can further expose 2-byte and 8-byte integers in the future if necessary. Lastly, since `i8` and `u8` is rarely used in practice, and to implement it would require us to use the user-define comparators in LMDB, I think we can leave them out.
> 
> Does it make sense? @mykmelez

Yes, thanks for the detailed explanation! It's really helpful to have someone with LMDB expertise looking at my patches. :-)

mykmelez:
@ncloudioj I've re-removed the `u16` and `u64` implementations, but I added newtype-based impls for those types to tests/integer-store.rs, to demonstrate how you could implement this for yourself (along with a warning about the danger of doing so). Does that seem reasonable?

ncloudioj:
@mykmelez Yep, this definitely hits the sweet spot of this particular feature. 👍

r=nanj. Go ahead merge it after fixing the rustfmt nit.

PR #100, opened 2018-12-20
To 2018!

rrichardson:
Updates to Cargo.toml and module use/import to be 2018 edition compliant.


mykmelez:
@rrichardson Looks good, just some `cargo fmt` errors, and I also noticed that test modules are still using `extern crate` declarations. I fixed both of those issues and requested a pull into your branch in https://github.com/rrichardson/rkv/pull/2.

PR #101, opened 2018-12-21
Refactor into store-centric API

rrichardson:
This depends on the 2018 edition PR #100 

The `Store` types are now 
```
SingleStore,
MultiStore,
IntegerStore,
MultiIntegerStore
```
The Reader/Writer types have been completely removed.. this now uses
```
lmdb::{ 
  Transaction, 
   RwTransaction, 
   RoTransaction, 
}
```
all `*Store::get` methods take a `T: Transaction` instance. 

1 additional interface change: 
1. When creating from Rkv::open... if we actually made a method for every permutation of store-type, flags, and create/not we'd have 16 create* functions..   So I created 4 open* functions which all take a `open : bool` and a `flags : DatabaseFlags`  which all do what you'd expect.


rrichardson:
OK.. @ncloudioj  @mykmelez  -  This should be good to go.   When fixing up the tests, I'm pretty happy how the types and lifetimes got cleaned up a little. 

`IntegerStore` now will only work against a single type of PrimitiveInt (you can still open it with the wrong type) 

Also, check out the Rkv::open* methods,  I don't having to pass the extra params.. I guess the other option is to create a StoreConfig to house the flags and create-vs-open option.

mykmelez:
@rrichardson, this branch could use a merge from master as well. Currently it still shows changes that have landed on master in #100.

rrichardson:
@mykmelez  - huh.. git tells me that my store-api-refactor branch is up to date with master.

rrichardson:
also says its up to date with rrichardson:2018-edition which merged in your changes..

mykmelez:
Hmm, strange. The "Files changed" tab of this pull request is still showing the changes from #100. For example, examples/iterator.rs shows replacement of `extern crate rkv;` with `use rkv;`, even though those changes have landed on master.

mykmelez:
>[Merge branch 'master' of https://github.com/mozilla/rkv into store-api-refactor](/mozilla/rkv/pull/101/commits/a939dcda2ff627059e50ad8a229a86b8b950d128)

Thanks, that seems to have worked! The "Files changed" tab now only shows the changes on this branch, which makes it easier to review!

src/store/integermulti.rs
```diff
@@ -0,0 +1,117 @@
...
+        macro_rules! test_integer_keys {
+            ($type:ty, $key:expr) => {{
+                let mut writer = k.write().expect("writer");
+
+                s.put(&mut writer, $key, &Value::Str("hello!")).expect("write");
+                assert_eq!(s.get_first(&writer, $key).expect("read"), Some(Value::Str("hello!")));
+                writer.commit().expect("committed");
+
+                let reader = k.read().expect("reader");
+                assert_eq!(s.get_first(&reader, $key).expect("read"), Some(Value::Str("hello!")));
```
mykmelez: This could use a test of reading multiple values for a key from an integermulti store (either here or in env.rs).

src/store/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod integer;
+pub mod integermulti;
+pub mod multi;
+pub mod single;
```
mykmelez: Nit: per [No more mod.rs](https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs), this file should be located at _src/store.rs_.

src/store/integermulti.rs
rrichardson: Done.

rrichardson:
Arg.. I just ran `cargo +nightly fmt` and it still finds differences..

src/lib.rs
```diff
@@ -24,8 +25,11 @@
 //! - [Manager](struct.Manager.html): a singleton that controls access to LMDB environments
 //! - [Rkv](struct.Rkv.html): an LMDB environment, which contains a set of key/value databases
 //! - [Store](struct.Store.html): an LMDB database, which contains a set of key/value pairs
+//! - [Transaction](https://github.com/mozilla/lmdb-rs/blob/master/src/transaction.rs#L16): A
+//! ReadOnly or Writeable transaction and a Transaction trait which abstracts over them.
```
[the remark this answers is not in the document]
mykmelez: Sounds good.

Eventually I'd like to wrap the lmdb-rs transaction types in abstractions, as we do for lmdb-rs databases, to make it easier for a consumer to understand the rkv API (without having to read documentation of its lmdb-rs dependency).

But that's orthogonal to these changes and should happen separately anyway.

src/store/multi.rs
```diff
@@ -0,0 +1,161 @@
...
+        txn: &'env T,
+        k: K,
+    ) -> Result<Option<Value<'env>>, StoreError> {
+        let result = txn.get(self.db, &k);
+        read_transform(result)
+    }
+
+    /// Insert a value at the specified key.
+    /// This put will allow duplicate entries.  If you wish to have duplicate entries
+    /// rejected, use the `put_with_flags` function and specify NO_DUP_DATA
```
mykmelez: Nit: we should add tests for the _put_with_flags()_ functions as well, although we can do that in a followup.

src/store/multi.rs
```diff
@@ -0,0 +1,161 @@
...
+        Ok(MultiIter {
+            iter,
+            cursor,
+        })
+    }
+    */
+}
+
+/*
+impl<'env> Iterator for MultiIter<'env> {
```
mykmelez: Nit: this comment block (and the next one) references a _MultiIter_ type, but that type isn't defined anywhere (not even in another comment block). It'd be helpful to include a declaration of that type along with this partial implementation, even if it's just a skeletal one. Or, perhaps better, move this code into an issue that tracks the problem and its resolution.

src/store/single.rs
```diff
@@ -0,0 +1,116 @@
...
+    Transaction,
+};
+
+use lmdb::WriteFlags;
+
+use crate::error::StoreError;
+
+use crate::value::Value;
+
+fn read_transform(val: Result<&[u8], lmdb::Error>) -> Result<Option<Value>, StoreError> {
```
mykmelez: Nit: the two implementations of _read_transform()_ in single.rs and multi.rs are identical and could be refactored into their super module store.rs.

Cargo.toml
```diff
@@ -18,8 +18,8 @@ backtrace = ["failure/backtrace", "failure/std"]
 [dependencies]
 arrayref = "0.3"
 bincode = "1.0"
+lmdb-rkv = "0.11"
 lazy_static = "1.0.2"
-lmdb-rkv = "0.9"
```
mykmelez: Nit: I'd preserve the alphabetical ordering here.

src/store/multi.rs
```diff
@@ -0,0 +1,161 @@
...
+
+fn read_transform(val: Result<&[u8], lmdb::Error>) -> Result<Option<Value>, StoreError> {
+    match val {
+        Ok(bytes) => Value::from_tagged_slice(bytes).map(Some).map_err(StoreError::DataError),
+        Err(lmdb::Error::NotFound) => Ok(None),
+        Err(e) => Err(StoreError::LmdbError(e)),
+    }
+}
+
+fn read_transform_owned(val: Result<&[u8], lmdb::Error>) -> Result<Option<OwnedValue>, StoreError> {
```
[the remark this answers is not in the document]
rrichardson: I think this was related to the MultiIter impl that I mostly removed because I failed to implement it.

src/store/multi.rs
```diff
@@ -0,0 +1,161 @@
...
+*/
+
+impl<'env> Iterator for Iter<'env> {
+    type Item = (&'env [u8], Result<Option<Value<'env>>, StoreError>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.iter.next() {
+            None => None,
+            Some(Ok((key, bytes))) => Some((key, read_transform(Ok(bytes)))),
+            Some(Err(_)) => None,
```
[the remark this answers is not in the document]
rrichardson: 👍

PR #105, opened 2019-01-12
update lmdb-rkv to latest version 0.11

mykmelez:
@ncloudioj Per https://github.com/mozilla/lmdb-rs/pull/14, the latest version of the lmdb-rkv crate returns an error instead of panicking when `mdb_cursor_get()` returns an error. That changes the return value of _lmdb::Iter.next()_, which is now a _Result_; so we need to update _rkv::Iter.next()_ accordingly.

We could maintain backward compatibility by continuing to return `(&'env [u8], Result<Option<Value<'env>>, StoreError>)` from _rkv::Iter.next()_, either ignoring an _Err_ result from _lmdb::Iter.next()_ or hiding it in the key-value tuple's _value_ field (which can currently be a _StoreError_ if _read_transform()_ fails to transform the value). But that's either a silent failure or a misleading result.

Alternately, we could distinguish between a  _StoreError_ result of _lmdb::Iter.next()_ and a _StoreError_ result of _read_transform()_ by returning `Result<(&'env [u8], Result<Option<Value<'env>>, StoreError>), StoreError>`, i.e. either an error or a key-value tuple whose _value_ field may contain an error. But that feels overcomplex, requiring two layers of error handling by the consumer.

This branch adopts the third approach of reporting all errors while consolidating error reporting by returning `Result<(&'env [u8], Option<Value<'env>>), StoreError>`, i.e. either a key-value tuple (that does not contain an error) or a _StoreError_ (which can be either from _lmdb::Iter.next()_ or from _read_transform()_).

As this is a breaking change, the branch also increments the minor version of rkv.


mykmelez:
> I'd assume it's unlikely to encounter `EINVAL` in the course of iteration.

Per this [code comment](https://github.com/danburkert/lmdb-rs/blob/c64f0b4f8280190525a6430388e714739f18976b/src/cursor.rs#L250-L251), it actually can encounter `EINVAL` in the course of iteration, if the cursor was previously seeked to a non-existent value, as when you call `Cursor.iter_from()` with a key greater than all keys in the database.

Perhaps there's a cheap way to detect that before returning the iterator, unsure. However, note that the documentation for _mdb_cursor_get()_ describes `MDB_NOTFOUND` and `EINVAL` as "some possible errors," so there may be others that aren't listed but may be returned. Thus I'm not sure how feasible it is to avoid error results from _Iter.next()_. But it's certainly worth investigating!

ncloudioj:
> Per this code comment, it actually can encounter EINVAL in the course of iteration, if the cursor was previously seeked to a non-existent value, as when you call Cursor.iter_from() with a key greater than all keys in the database.

Right, I was wondering if we could capture those EINVALs in the iterator creators, before returning them to the consumers. e.g. if calling iter_from() with a key greater than all keys,  then returns an `Err(...)`, otherwise, returns an `OK(iter)`. So iterators in rkv always start from the valid state.

After taking a quick glimpse of mdb_cursor_get, it indeed only exposes `MDB_NOTFOUND` and `EINVAL`, which acts as a general error for any the cursor ops they believe being invalid.

Will take a deeper look into this, stay stuned! :)

Issue #49, opened 2018-06-21
document various limitations of LMDB


ncloudioj:
We might want to document various limitations of LMDB in order to offer the "least surprise" for the rkv users. Off top of my head, LMDB has following limitations:

* Max key size (`MDB_MAXKEYSIZE = 511 bytes`), it also applies to the value for the dupsort store. Note that it's a compile-time configuration, and can't be changed at runtime
* Max environment size (default as 10 MB), once the environment is full, following writes will fail
* Max number of database (default as 5), hosting a moderate number (say up to a few dozens) of databases in a single environment is fine, however, hosting too many has both memory and performance impact
* Max readers (default as 126), the maximum readers/threads allowed to access an LMDB environment. Passing this limit will end up failing to create readers
* Too many writes in a single transaction. Unsure what exactly the maximum writes is, but LMDB may complain while conducting a bulk load write in a single transaction. To workaround this, we can periodically commit a transaction and create a new one for the rest of writes.

ncloudioj:
* Writes are serialized by a write lock, which means that there is only one active writer allowed at any point of time, and other writers will be blocked until that active writer aborts/commits. As such, favor the short write transactions in the multi-writer scenario, also mind the blocking for those inactive writers.

mykmelez:
> Max environment size (default as 10 MB), once the environment is full, following writes will fail

The LMDB docs say that the default is 10MB, but the code actually sets DEFAULT_MAPSIZE to 1MB:

https://github.com/mozilla/lmdb/blob/4f9fe9fcead4ce3c49d80d39526472b5c274b188/libraries/liblmdb/mdb.c#L729

Another limitation it would be useful to document is the maximum size of a value. Presumably it's the [MAXDATASIZE](http://www.lmdb.tech/doc/group__internal.html#ga16c16f9369be4a374a3e621f6d13bb16) of 0xffffffffUL (4,294,967,295).

PR #107, opened 2019-01-18
remove mut from the *store mutation functions as it is not needed

rrichardson:
Fixes the *Store put/del functions that took &mut self.  Changes everything to &self and removes the warnings from all code that consumes those methods.

### DIFF
--- a/examples/iterator.rs
+++ b/examples/iterator.rs
@@ -59,7 +59,7 @@ fn main() {
     }
 }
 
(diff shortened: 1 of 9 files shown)


mykmelez:
NB: see https://github.com/mozilla/rkv/pull/16 for related discussion.

ncloudioj:
Thanks @mykmelez for pointing it out.

Indeed, Rust compiler can't completely capture this since the contending writer could be from another process. Added this note to #49.

PR #108, opened 2019-01-23
integrate clippy into CI

### LINKED ISSUE #68: integrate clippy in some fashion
Over in #65, @ncloudioj fixed some clippy nits. We should consider integrating clippy in some fashion to reduce the risk of introducing more such nits.

Per https://github.com/rust-lang-nursery/rust-clippy, "Since this is a tool for helping the developer of a library or application write better code, it is recommended not to include Clippy as a hard dependency. Options include using it as an optional dependency, as a cargo subcommand, or as an included feature during build. These options are detailed below."

I'm unsure which of these options is best.

ncloudioj:
This closes #68.

With this change, CI would fail if Clippy sees any warning in the patch, picky...picky ;)

Also silenced a few Clippy warnings (`passing by value instead of ref` and `naming convention`) in this patch, assuming that we do want to keep those behaviors for now.

### DIFF
--- a/src/store/multi.rs
+++ b/src/store/multi.rs
@@ -72,11 +72,7 @@ impl MultiStore {
     }
 
     /// Provides a cursor to all of the values for the duplicate entries that match this key
-    pub fn get_first<T: Transaction, K: AsRef<[u8]>>(
-        self,
-        txn: &T,
-        k: K,
-    ) -> Result<Option<Value>, StoreError> {
+    pub fn get_first<T: Transaction, K: AsRef<[u8]>>(self, txn: &T, k: K) -> Result<Option<Value>, StoreError> {
         let result = txn.get(self.db, &k);
         read_transform(result)
     }
--- a/src/store/multi.rs
+++ b/src/store/multi.rs
@@ -8,8 +8,6 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
-#![allow(clippy::trivially_copy_pass_by_ref)]
-
 use lmdb;
 
 use lmdb::{
@@ -64,7 +62,7 @@ impl MultiStore {
     }
 
     /// Provides a cursor to all of the values for the duplicate entries that match this key
-    pub fn get<'env, T: Transaction, K: AsRef<[u8]>>(&self, txn: &'env T, k: K) -> Result<Iter<'env>, StoreError> {
+    pub fn get<T: Transaction, K: AsRef<[u8]>>(self, txn: &T, k: K) -> Result<Iter, StoreError> {
         let mut cursor = txn.open_ro_cursor(self.db).map_err(StoreError::LmdbError)?;
         let iter = cursor.iter_dup_of(k);
         Ok(Iter {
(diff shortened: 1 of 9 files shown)


ncloudioj:
Whew, finally Travis is happy now.

r? @mykmelez

mykmelez:
Thanks, @ncloudioj, this is really helpful.

I see a few more warnings when I run Clippy (version `0.0.212 (2e26fdc2 2018-11-22)`) locally:

```
warning: methods called `new` usually return `Self`                                                                                                                  
  --> src/env.rs:58:5                                                                                                                                                
   |                                                                                                                                                                 
58 | /     pub fn new(path: &Path) -> Result<Rkv, StoreError> {                                                                                                      
59 | |         Rkv::with_capacity(path, DEFAULT_MAX_DBS)                                                                                                             
60 | |     }                                                                                                                                                         
   | |_____^                                                                                                                                                         
   |                                                                                                                                                                 
   = note: #[warn(clippy::new_ret_no_self)] on by default                                                                                                            
   = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/master/index.html#new_ret_no_self                                           
                                                                                                                                                                     
warning: methods called `new` usually return `Self`                                                                                                                  
  --> src/store/integer.rs:68:5                                                                                                                                      
   |                                                                                                                                                                 
68 | /     pub(crate) fn new(k: K) -> Result<Key<K>, DataError> {                                                                                                    
69 | |         Ok(Key {                                                                                                                                              
70 | |             bytes: k.to_bytes()?,                                                                                                                             
71 | |             phantom: PhantomData,                                                                                                                             
72 | |         })                                                                                                                                                    
73 | |     }                                                                                                                                                         
   | |_____^                                                                                                                                                         
   |                                                                                                                                                                 
   = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/master/index.html#new_ret_no_self                                           
                                                                                                                                                                     
warning: this argument is passed by value, but not consumed in the function body                                                                                     
  --> src/store/integer.rs:68:26                                                                                                                                     
   |                                                                                                                                                                 
68 |     pub(crate) fn new(k: K) -> Result<Key<K>, DataError> {                                                                                                      
   |                          ^ help: consider taking a reference instead: `&K`                                                                                      
   |                                                                                                                                                                 
   = note: #[warn(clippy::needless_pass_by_value)] on by default                                                                                                     
   = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/master/index.html#needless_pass_by_value                                    
                                                                                                                                                                     
    Finished dev [unoptimized + debuginfo] target(s) in 30.53s
```

We should probably squelch the `clippy::new_ret_no_self` warnings, as we have to return a `Result` from those functions, because they're fallible. (I suppose we could rename them instead.) We can presumably fix the `clippy::needless_pass_by_value` warnings by passing a reference instead.

src/store/multi.rs
```diff
@@ -8,6 +8,8 @@
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
 
+#![allow(clippy::trivially_copy_pass_by_ref)]
```
mykmelez: Perhaps we should fix these instead of suppressing the warning. Any reason not to do so?

ncloudioj:
> I see a few more warnings when I run Clippy (version 0.0.212 (2e26fdc2 2018-11-22)) locally

Yes, those appear to be legit warnings, but somehow Clippy didn't report here. my Clippy version is (clippy 0.0.212 (280069dd 2019-01-22)). Maybe the Rust version matters (mine is on the latest nightly)?

src/store/multi.rs
ncloudioj: Good question :)

I recall that I've fixed them in https://github.com/mozilla/rkv/pull/88, but those stubborn references came back again perhaps with the recent commits.

Let's fix them again.

ncloudioj:
@mykmelez No worries.

I've also run into this issue before for both rustfmt and clippy, I fixed it as follows:

```sh
$ rustup component remove rustfmt-preview
$ rustup component remove rustfmt-preview
$ rustup component add rustfmt
$ rustup component add clippy
```

mykmelez:
It looks like in my case the issue is that clippy is actually unavailable for today's macOS nightly, per https://mexus.github.io/rustup-components-history/x86_64-apple-darwin.html.

PR #111, opened 2019-01-28
fix nits and docs

mykmelez:
This fixes a couple of the nits noted in #101. It also fixes the crate-level documentation in src/lib.rs, which contains some outdated references to documentation pages for the _Store_ and _IntegerStore_ types.

### DIFF
--- a/src/store/multi.rs
+++ b/src/store/multi.rs
@@ -51,7 +51,7 @@ impl MultiStore {
         })
     }
 
-    /// Provides a cursor to all of the values for the duplicate entries that match this key
+    /// Provides the first value that matches this key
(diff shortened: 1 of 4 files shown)


src/store/multi.rs
```diff
@@ -129,6 +109,14 @@ impl MultiStore {
 }
 
 /*
+fn read_transform_owned(val: Result<&[u8], lmdb::Error>) -> Result<Option<OwnedValue>, StoreError> {
```
ncloudioj: Why added this comment here?

ncloudioj: Ah, I see. Let's keep it then. Thanks!

PR #123, opened 2019-02-14
implement From<OwnedValue> for Value

mykmelez:
Over in https://phabricator.services.mozilla.com/D19436#inline-106713, @mystor wondered why rkv doesn't provide a conversion from `OwnedValue` to `Value`, which is indeed a missing piece in rkv's API that requires downstream consumers like kvstore to implement the conversion themselves.

Here's an implementation that enables downstream consumers like kvstore to avoid implementing it themselves. Along the way, I removed unnecessary `ref` annotations now that rkv specifies `edition = "2018"` and thus [infers the binding mode for match bindings](https://rust-lang-nursery.github.io/edition-guide/rust-2018/ownership-and-lifetimes/default-match-bindings.html).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,16 @@ matrix:
     - rust: nightly
   fast_finish: true
 before_script:
-  - rustup toolchain install nightly
-  - rustup component add rustfmt-preview --toolchain nightly
+  # We install a known-to-have-rustfmt version of the nightly toolchain
+  # in order to run the nightly version of rustfmt, which supports rules
+  # that we depend upon.
+  - rustup toolchain install nightly-2019-02-07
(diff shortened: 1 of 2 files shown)


mykmelez:
The build failure is because the latest nightly version of Rust doesn't have rustfmt. It looks like we'll have to go back to pinning a known-to-have-rustfmt version of the nightly toolchain in order to ensure that we can run rustfmt. I did that in 3762a8f.

mykmelez:
FWIW, the Rust nightly we use could be almost anything other than today's nightly, as at least the last week's worth of nightly builds (besides today's) all have rustfmt, according to https://mexus.github.io/rustup-components-history/x86_64-apple-darwin.html.

PR #129, opened 2019-03-12
test that one can mutate managed Rkv instances in surprising ways

mykmelez:
Because Manager stores Rkv instances as Arc<RwLock<Rkv>>, and RwLock has interior mutability, it's possible to replace a managed Rkv instance, resulting in surprising behavior when retrieving instances from the Manager.

This branch just demonstrates the surprising behavior of Manager when one of its mapped Rkv instances is replaced by a new instance.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -143,6 +143,39 @@ mod tests {
         assert!(Arc::ptr_eq(&created_arc, &fetched_arc));
     }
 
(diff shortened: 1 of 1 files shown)


victorporof:
Can we land this? @ncloudioj

ncloudioj:
I think so, I will check it in.

PR #131, opened 2019-03-22
Export  'Readable' trait to allow user making generic method that take '&Readable' as parameter instead of '&Read' or '&Write'

asyade:
Everything is in the title, my approach is right or there is another one that i have not seen ?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -222,6 +222,7 @@ pub use lmdb::{
 };
 
 pub use self::readwrite::{
+    Readable,
     Reader,
     Writer,
 };


ncloudioj:
LGTM!

@mykmelez Is there any reason to not export this trait?

mykmelez:
I don't have any objection in principle, but I'd like to understand better how the trait would be used if it was public.

It isn't actually possible to make a function with a `&Readable` trait object parameter:

```rust
fn foo(readable: &Readable) {
    …
}
```

Because "error[E0038]: the trait `rkv::readwrite::Readable` cannot be made into an object… note: method `get` has generic type parameters".

It would be possible to use `Readable` as a trait bound:

```rust
fn bar<T: Readable>(readable: &T) {
    …
}
```

But both of its methods—`get()` and `open_ro_cursor()`—take a `Database`. And while the `SingleStore` and `MultiStore` structs have a `db: Database` field, that field is private. So it would only be possible to use the trait bound with your own `Database`.

@asyade Can you explain how you'd use `Readable`, if it were made public? It can just be in a comment here, but if you add a code example to the examples/ directory as part of your patch, then we can ensure that we don't break your usage in the future!

asyade:
I will add some example in the next days, (i'm using the generic type parametter syntaxe sorry for the mistake), if i used this approach its beacause my function that take a Readable as parameter use multiple store transaction and obviously i would make only one reader/writer by transaction

rrichardson:
I assumed that `Readable` was omitted by accident, which is why I made #145 .  I should have checked the PRs first.  I have a fair amount of code in Megadex  that expects access to `Readable`. 

An example for how I use transactions is here: 
https://github.com/rrichardson/megadex/blob/master/megadex_rkv/src/lib.rs#L187-L188

rrichardson:
@mykmelez  - Is my explanation intelligible?  Can we get your blessing here?

mykmelez:
@rrichardson I must be missing something, since you and @asyade are clearly running into a limitation that can be addressed by making `Readable` public; but I still don't understand the limitation.

I tried to convert your code snippet into a test function, and I haven't been able to design one that fails because `Readable` is private. Here's what I have currently:

```rust
#[test]
fn test_multi_store_transaction() {
    let root = Builder::new().prefix("test_multi_store_transaction").tempdir().expect("tempdir");
    fs::create_dir_all(root.path()).expect("dir created");
    let rkv = Rkv::with_capacity(root.path(), 2).expect("rkv");
    let store1 = rkv.open_single("store1", StoreOptions::create()).expect("opened");
    let store2 = rkv.open_single("store2", StoreOptions::create()).expect("opened");

    // First write a key/value pair to each store, where the value
    // in the first store is the key in the second store.
    {
        let mut writer = rkv.write().expect("writer");
        store1.put(&mut writer, "store1-key", &Value::Str("store2-key")).expect("wrote");
        store2.put(&mut writer, "store2-key", &Value::Str("store2-val")).expect("wrote");
        writer.commit().expect("committed");
    }

    // Then use a single Reader to get the values from the two stores.
    let reader = rkv.read().expect("reader");
    let store2_key = match store1.get(&reader, "store1-key").expect("read") {
        Some(Value::Str(value)) => value.to_owned(),
        _ => panic!("unexpected"),
    };
    let store2_val = store2.get(&reader, store2_key).expect("read");
    assert_eq!(store2_val, Some(Value::Str("store2-val")));
}
```

Can you modify this to fail because `Readable` is private? Alternately, could you provide a small program for the examples/ directory that demonstrates the failure?

I'm open to making `Readable` public. I just want to understand the use case for it, for which a working code sample (whether a test or an example) would be very helpful.

rrichardson:
@mykmelez  -  My understanding is that `Readable` is required to define functions which are meant to be run within a transaction. 
So making a single "test function" wouldn't run into the problem.. but making a "test three functions" would.  :) 

https://github.com/rrichardson/megadex/blob/master/megadex_rkv/src/lib.rs#L131-L189

Is an example.  I'll create a PR for a simpler test case.

rrichardson:
OK.  #146 demonstrates the need for a public Readable as I understand it.

mykmelez:
> OK. #146 demonstrates the need for a public Readable as I understand it.

Thanks, I see it now! I'm sorry for being so slow to understand this. I'll merge this branch, and then if you merge master into #146, I'll merge that branch once the tests pass there.

PR #132, opened 2019-03-29
WIP - auto resizing

ncloudioj:
Took a crack at the auto-resizing feature.

I've made following changes:
* Added a new lock for the resize coordination to the Rkv struct.
* Augmented the `reader&writer` structure to hold the RwLockGuards, so the resize can not be done if there are active transactions in that process.
* Added a function `set_map_size()` to Rkv for resizing
 
Caveats:
* Deadlock: if the consumer calls `set_map_size()` in the same thread if there is any active transactions
* It only coordinates among rkv readers and writers, but not the raw `lmdb::Ro(w)Traction`s. For example, it doesn't consider those implicit transactions used when opening/creating stores.

Assuming that eventually we want to do the (auto) resizing:
* If rkv detects that the environment is "almost" full (perhaps when the consumer firstly opens the write transaction?).
* Or if the consumer encounters the "MapFull" errors.  

@mykmelez Would like to hear your thoughts first before I go any further!


mykmelez:
> Took a crack at the auto-resizing feature.
> 
> I've made following changes:
> 
> * Added a new lock for the resize coordination to the Rkv struct.
> 
> * Augmented the `reader&writer` structure to hold the RwLockGuards, so the resize can not be done if there are active transactions in that process.
> 
> * Added a function `set_map_size()` to Rkv for resizing

That seems like a straightforward way to implement this feature.

> Caveats:
> 
> * Deadlock: if the consumer calls `set_map_size()` in the same thread if there is any active transactions

This is presumably similar to the one that occurs when you create a new store while a write transaction is active. In the future, we should find a way to avoid this deadlock, perhaps by returning an error when the consumer attempts to perform an operation that uses a write transaction internally while another write transaction is active.

> * It only coordinates among rkv readers and writers, but not the raw `lmdb::Ro(w)Traction`s.

This seems reasonable, since the goal is to support auto-resizing for consumers using the high-level rkv APIs. Consumers who use the low-level LMDB APIs are also presumably capable of managing their own map size.

> For example, it doesn't consider those implicit transactions used when opening/creating stores.

These are a different category, since they're high-level operations that happen to use write transactions internally. rkv should handle `MapFull` errors when creating a store; but is it even possible to encounter such errors in that case?

> Assuming that eventually we want to do the (auto) resizing:
> 
> * If rkv detects that the environment is "almost" full (perhaps when the consumer firstly opens the write transaction?).

Yes, provided that we can find a reasonable heuristic for determining when to increase the map size (and by how much).

> * Or if the consumer encounters the "MapFull" errors.

We should definitely auto-resize in this case.

ncloudioj:
@mykmelez Thanks for your insights!

Let's now take a look at how to deal with the MapFull error. Turns out it's more involved than what I've expected. The main complexity is from the transaction recovery after the resize, since the resizing requires all the transactions to end first.

Consider the following case:

```rs
let writer = rkv.write();
store.put(writer, "key", "value", writeflag); // succeeded
store.put(writer, "key1", "value1", writeflag); // succeeded
store.put(writer, "key2", "value2", writeflag); // failed: MapFull
...
// The consumer may commit or abort the writer later on.
```

To handle the `MapFull` here, presumably it needs to: 1) abort the current writer (all uncommitted writes will be rolled back too); 2) increase the map size; 3) open a new write transaction; 4) redo all the previous writes (put, clear, delete).

Managed to crank out a draft implementation for this in b098ed4b0fe12ac1bd45f23e36078fd7c715e90f. NB: I've added a new struct `WriterEx` instead of modifying the `Writer` so it's easier to see the difference. 

Some obvious drawbacks:
* The transaction recovery won't be 100% reliable, particularly for the multi-process scenarios. Over the course of resizing, another process may have committed some changes to the same store, those succeeded writes in the prior-resizing transaction might fail in the post-resizing transaction.
* The memory overhead of the redo log, which can be unsired for the bulk insert transactions.
* The code complexity.

@mykmelez Would like to hear your thoughts again:
* Have I over-complicated it?
* Do you have any better ideas?

mykmelez:
@ncloudioj and I chatted about this, and the gist is that he hasn't overcomplicated this, as it's inherently very complicated to auto-resize the map while maintaining the API's existing promises (atomic transactions, writers in multiple processes, etc.).

Our consensus is that we should take a step back and focus on improving the handling of MapFull errors without auto-resizing during a write transaction. Specifically:

1. Investigate auto-resizing of the map when initially opening an environment, if the database is almost full (by some definition of full), in order to reduce the likelihood of MapFull errors on write.
2. Provide examples/tests that demonstrate the best practices for a consumer to handle this error itself (f.e. by aborting the transaction, resizing the map, and then repeating the transaction).

ivzhh:
If you don't mind, I hope to add some LMDB API context to this pull request for documentation. 

If call `mdb_put()` and receive `MDB_MAP_FULL`, it shows the DB is full. If call `mdb_put()` again, the return code is `MDB_BAD_TXN`. So this transaction cannot be reused. The `mdb_env_set_mapsize()` can only be called after this broken transaction being aborted.

> This function should be called after mdb_env_create() and before mdb_env_open(). It may be called at later times if no transactions are active in this process.

ncloudioj:
@ivzhh Hi, this is a WIP patch attempting to implement the auto-resizing feature for the MapFull error. I've created a separate PR #135, which has better [documentation](https://github.com/mozilla/rkv/pull/135/files#diff-06aeb1dbfcf2eb1e00a4f7fe0edab250R238), to add the `set_map_size()`. 

> If call mdb_put() and receive MDB_MAP_FULL, it shows the DB is full. If call mdb_put() again, the return code is MDB_BAD_TXN. So this transaction cannot be reused. The mdb_env_set_mapsize() can only be called after this broken transaction being aborted.

Correct, any further actions on a MapFull'ed transaction will fail. Currently rkv doesn't handle this particular error for the consumers, so they needs to explicitly capture it and decide how to handle it. Aborting the transaction is usually the first step of error handling.

ivzhh:
@ncloudioj Hi. Thank you for your detailed reply! I totally agree with you. I added that comment because I was wondering what will happened if `MDB_MAP_FULL` returned. I add that for documentation when people may search about that.

I read current pull request and I agree with your changes. For sure we need to maintain a redo log for any commit failure. Persistence is only guaranteed after `0 == mdb_txn_commit()`. So I second your design.

leo-yuriev:
Take look to the https://github.com/leo-yuriev/libmdbx#improvements-over-lmdb

ncloudioj:
> Take look to the https://github.com/leo-yuriev/libmdbx#improvements-over-lmdb

Yes, we're aware of this project, and it does provides some great insights.

For the auto resizing, it is one of libmdbx's design goals, and has been implemented directly in the storage engine. Whereas rkv focus on improving the ergonomics of LMDB, instead of modifying the storage engine, we are looking to solve it from the library consumers standpoint of view. Hence the solution that works in libmdbx maybe not applicable here.  

FWIW, we've added various helper functions (`rkv::set_map_size()`, `rkv::load_ratio()`) to facilitate the map re-sizing in the 0.9.5 release. It's still not auto-resizing though, and it's up to the consumers to enforce the resizing conditions. We will keep an eye on other improvements in order to make the rkv consumers life easier.

ncloudioj:
Close it as this turned out to be non-trivial to implement in rkv. We have provided various auto-resizing APIs for the consumers to do so at the application level.

PR #135, opened 2019-04-03
Add set_map_size and info for Rkv

ncloudioj:
This is a spin-off of #132, which includes:
* `Rkv.info()`
* `Rkv.set_map_size()`
* `Rkv.load_ratio()`


src/env.rs
```diff
@@ -791,6 +847,65 @@ mod tests {
...
+
+        // The default size is 1MB.
+        assert_eq!(info.map_size(), 1024 * 1024);
+        // Should greater than 0 after the write txn.
+        assert!(info.last_pgno() > 0);
+        // A txn to open_single + a txn to write.
+        assert_eq!(info.last_txnid(), 2);
+        // The default max readers is 126.
+        assert_eq!(info.max_readers(), 126);
+        assert_eq!(info.num_readers(), 0);
```
mykmelez: Is it possible to test that this number increments if we create a read transaction?

src/env.rs
```diff
@@ -791,6 +847,65 @@ mod tests {
...
+        let root = Builder::new().prefix("test_load_ratio").tempdir().expect("tempdir");
+        fs::create_dir_all(root.path()).expect("dir created");
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        let sk: SingleStore = k.open_single("sk", StoreOptions::create()).expect("opened");
+        let mut writer = k.write().expect("writer");
+        sk.put(&mut writer, "foo", &Value::Str("bar")).expect("wrote");
+        writer.commit().expect("commited");
+
+        let ratio = k.load_ratio().expect("ratio");
+        assert!(ratio > 0.0_f32 && ratio < 1.0_f32);
```
mykmelez: It'd also be interesting to test that the ratio increases after you write to the database.

src/env.rs
```diff
@@ -791,6 +847,65 @@ mod tests {
...
+
+    #[test]
+    fn test_set_map_size() {
+        let root = Builder::new().prefix("test_size_map_size").tempdir().expect("tempdir");
+        fs::create_dir_all(root.path()).expect("dir created");
+        let k = Rkv::new(root.path()).expect("new succeeded");
+        let sk: SingleStore = k.open_single("sk", StoreOptions::create()).expect("opened");
+
+        // The default size is 1MB.
+        const DEFAULT_SIZE: usize = 1024 * 1024;
```
mykmelez: This could be factored out to a static, since the same value is recalculated in `test_info()`.

ncloudioj:
Thanks!

Review feedback is now fixed.

PR #137, opened 2019-05-07
Consider freelist when calculating load ratio

ncloudioj:


ncloudioj:
Looks like Clippy introduced a new rule `cognitive_complexity` to replace the `cyclomatic_compleixity`, which complicated this check in stable and nightly.

It's time to break that function into several smaller ones to make this check happy :)

src/env.rs
```diff
@@ -259,7 +265,7 @@ impl Rkv {
     }
 }
 
-#[allow(clippy::cyclomatic_complexity)]
+#[allow(clippy::cognitive_complexity)]
```
mykmelez: It looks like this change breaks stable Rust. Can we allow both?

src/manager.rs
```diff
@@ -37,8 +37,8 @@ use crate::error::StoreError;
 
 use crate::Rkv;
 
-/// A process is only permitted to have one open handle to each Rkv environment.
-/// This manager exists to enforce that constraint: don't open environments directly.
+// A process is only permitted to have one open handle to each Rkv environment.
+// This manager exists to enforce that constraint: don't open environments directly.
```
mykmelez: I think the intent here was to show this comment in the docs for Manager, although it doesn't appear anywhere in the docs when it's placed at this location. Perhaps, instead of changing it from a doc comment, we should move it to the `struct Manager` declaration.

src/env.rs
ncloudioj: I couldn't find a way to get it work both for stable and nightly. Let me see if it makes sense to break those functions down into smaller ones.

ncloudioj: I ended up swapping the `cyclomatic_complexity` out with a more general `complexity` rule so that it worked both on nightly and stable. We can switch it back once Clippy stable deprecates `cyclomatic_complexity`.

mykmelez:
Sounds good!

PR #139, opened 2019-05-16
Add serialized_size to Value enum

ncloudioj:
This can be used to inspect the size of `rkv::Value`, which in turn could be useful for map resizing.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -177,10 +177,7 @@ impl<'s> Value<'s> {
             Value::Str(v) => serialize(&(Type::Str.to_tag(), v)),
             Value::Json(v) => serialize(&(Type::Json.to_tag(), v)),
             Value::Blob(v) => serialize(&(Type::Blob.to_tag(), v)),
-            Value::Uuid(v) => {
-                // Processed above to avoid verbose duplication of error transforms.
-                serialize(&(Type::Uuid.to_tag(), v))
-            },
+            Value::Uuid(v) => serialize(&(Type::Uuid.to_tag(), v)),
         }
         .map_err(DataError::EncodingError)
     }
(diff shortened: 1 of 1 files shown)


src/value.rs
```diff
@@ -216,3 +235,37 @@ impl<'s> From<&'s OwnedValue> for Value<'s> {
...
+    fn test_value_serialized_size() {
+        // | Value enum    | tag: 1 byte   |     value_payload        |
+        // |----------------------------------------------------------|
+        // |   I64         |     1         |       8                  |
+        // |   U64         |     1         |       8                  |
+        // |   Bool        |     1         |       1                  |
+        // |   Instant     |     1         |       8                  |
+        // |   F64         |     1         |       8                  |
+        // |   Uuid        |     1         |       16                 |
+        // | Str/Blob/Json |     1         |(8: len + sizeof(payload))|
```
mykmelez: This is extremely helpful in understanding the sizes being asserted below. Thanks!

src/value.rs
```diff
@@ -183,6 +184,24 @@ impl<'s> Value<'s> {
...
+            Value::Bool(v) => serialized_size(&(Type::Bool.to_tag(), *v)),
+            Value::U64(v) => serialized_size(&(Type::U64.to_tag(), *v)),
+            Value::I64(v) => serialized_size(&(Type::I64.to_tag(), *v)),
+            Value::F64(v) => serialized_size(&(Type::F64.to_tag(), v.0)),
+            Value::Instant(v) => serialized_size(&(Type::Instant.to_tag(), *v)),
+            Value::Str(v) => serialized_size(&(Type::Str.to_tag(), v)),
+            Value::Json(v) => serialized_size(&(Type::Json.to_tag(), v)),
+            Value::Blob(v) => serialized_size(&(Type::Blob.to_tag(), v)),
+            Value::Uuid(v) => {
+                // Processed above to avoid verbose duplication of error transforms.
```
mykmelez: Erm, one nit: this comment, which is also in `to_bytes()`, appears to be incorrect for both `to_bytes()` and `serialized_size()`, neither of which actually processes `Type::Uuid` earlier in the function.

It looks like the comment comes from `from_type_and_data()`, which does process `Type::Uuid` separately (i.e. before the general match statement).

ncloudioj: Copy&pasta oops! I'll fix it up.

PR #145, opened 2019-06-24
exposing Readable trait to consumers

rrichardson:
Make the Readable trait public, so that users can use Reader and Writer transactions interchangeably for read operations.


rrichardson:
Closing in favor of #131

PR #146, opened 2019-06-28
added tests for Readable txns

rrichardson:
This is an integration test demonstrating the use of helper functions within transactions.


victorporof:
This looks quite useful! Just need to make sure rustfmt passes. I'll do this before merging unless you want to :)

rrichardson:
@victorporof  Thanks!

PR #149, opened 2019-07-24
Update the url crate to 2.0

SimonSapin:


mykmelez:
@SimonSapin Are there any particular benefits to the newer version, or perhaps this is part of a broader upgrade to version 2.0 across the Gecko dependency graph? I'm happy to do this, I just want to understand the motivation!

SimonSapin:
The main motivation for url 2.0 was having serde 1.x support directly in the crate. (Previously we had older serde, or through the url_serde crate which is not as convenient has limitations.) And removing some other obsolete functionality. For many users it won’t change much.

We don’t plan on making more url 1.x releases (though it’s possible if the need gets pressing). But we might want to use some new functionality in Servo at some future point. I’m upgrading in anticipation, so it’s not blocking then. We also try to avoid duplicate dependencies (multiple versions of the same crate) in Servo. One of the Servo dependencies that uses `url` is `webdriver`, which has its source in mozilla-central. https://bugzilla.mozilla.org/show_bug.cgi?id=1568540 is about upgrading it. To avoid the same duplication there, I also went through crates that use `url` and are used by Gecko.

mykmelez:
That makes sense! I'll merge this and publish it to crates.io.

mykmelez:
Published as https://crates.io/crates/rkv/0.9.7.

SimonSapin:
Thanks!

Issue #117, opened 2019-02-06
hide lmdb::RoCursor type from consumers of Reader/Writer structs


mykmelez:
The _Readable_ trait in readwrite.rs currently leaks the _lmdb::RoCursor_ type, and we generally don't want to show consumers the types from the lmdb crate dependency (preferring to encapsulate them in higher-level rkv types). We should figure out a way to hide that type.

PR #158, opened 2019-10-21
Implement multiple backing stores for Rkv

victorporof:
WIP for allowing Rkv to use multiple backing stores.
"Safe mode" backing store implementation should now be a breeze.

### DIFF
--- a/src/backend/traits.rs
+++ b/src/backend/traits.rs
@@ -23,7 +23,7 @@ use crate::error::StoreError;
 
 pub trait BackendError: Debug + Display + Into<StoreError> {}
 
-pub trait BackendDatabase: Debug + Eq + PartialEq {}
+pub trait BackendDatabase: Debug {}
 
 pub trait BackendFlags: Debug + Eq + PartialEq + Copy + Clone + Default {
     fn empty() -> Self;
--- a/src/backend/traits.rs
+++ b/src/backend/traits.rs
@@ -23,7 +23,7 @@ use crate::error::StoreError;
 
 pub trait BackendError: Debug + Display + Into<StoreError> {}
 
-pub trait BackendDatabase: Debug + Eq + PartialEq + Copy + Clone {}
+pub trait BackendDatabase: Debug + Eq + PartialEq {}
 
 pub trait BackendFlags: Debug + Eq + PartialEq + Copy + Clone + Default {
     fn empty() -> Self;
@@ -120,7 +120,7 @@ pub trait BackendRoTransaction: Debug {
     type Database: BackendDatabase;
     type Flags: BackendWriteFlags;
 
-    fn get(&self, db: Self::Database, key: &[u8]) -> Result<&[u8], Self::Error>;
+    fn get(&self, db: &Self::Database, key: &[u8]) -> Result<&[u8], Self::Error>;
 
     fn abort(self);
 }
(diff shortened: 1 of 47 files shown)


victorporof:
@ncloudioj For more context, see https://docs.google.com/document/d/1aoVEcn1PR-W8qE2wOx0FSchO8qMpqaAzbt4pHktf-Wc

victorporof:
Also fixes https://github.com/mozilla/rkv/issues/117

ncloudioj:
@victorporof could you please add some comments to outline the major changes of this patch? The attached document has given me a great holistic rationale for this change, it will be very helpful for me to reivew the code with those inputs.

Thanks!

victorporof:
@ncloudioj To recap, here's how I've split up this PR. It would be excellent if we could land this today.

#### 1. Pre: split key types into separate module

This commit makes it so that changeset no. 2 (a4d76eb) is readable in case of `src/store/integer.rs`.

#### 2. Generalize LMDB usage and add support for multiple backing stores

Isolates LMDB's surface area that this crate requires, splitting it into a set of interdependent traits. Structures implementing these traits are usable by RKV as a backing engine for key-value storage.

This commit also implements these traits for newtypes wrapping LMDB itself. This showcases the only changes required to user code when picking the LMDB backend, which involves the turbofish syntax (e.g `Rkv::new<Lmdb>(...)`).

Changeset no. 5 (4a955a7) actually implements the safe mode backend.

#### 3. Drop copy and clone traits requirement on backend databases

The current API requires that an environment returns an owned database (a.k.a. not a reference to one) to consumers when calling `open_db` or `create_db. This database type must be the same everywhere, a constraint expressed through the `Database` trait and associated types in the various traits defined in commit no. 2.

In case of LMDB, we do this by cloning when it comes to LMDB databases. Cloning is cheap in those situations because they're just wrappers around pointers to the C data structures.

In case of the safe mode implementation, we don't want to clone the entire database in this scenario. Returning an owned database without duplicating its contents is easy through smart pointers such as `Arc`. However, since Rust doesn't allow implementing traits on built-ins (a.k.a. can't impl `Database` for `Arc<_>`), and since we don't want to do this unsafely, dropping the Clone requirement is the way to go.

It might be possible to keep the clone requirement if we're not particular about having the `Database` trait, however I'd prefer to have strict requirements for the traits describing a backing store.

#### 4. Drop eq and partial-eq trait requirement on backend databases

Same as above (commit no. 3). A smart pointer such as `Arc` cannot derive equality.

#### 5. Implement safe mode backing store

Implements the traits describing a backing store (defined in commit no. 2). We don't aim for feature completeness here, just enough so that various RKV tests still pass (to verify that the implementation actually behaves like LMDB in the most common cases). CRLite support dictates how relaxed we can be about performance.

#### 6. Add more tests for the rkv backend and 7. Add more tests for the safe mode backend

These commits add more tests.

src/backend/traits.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    fn freelist(&self) -> Result<usize, Self::Error>;
+
+    fn set_map_size(&self, size: usize) -> Result<(), Self::Error>;
+}
+
+pub trait BackendRoTransaction: Debug {
+    type Error: BackendError;
+    type Database: BackendDatabase;
+    type Flags: BackendWriteFlags;
```
ncloudioj: Read-only transactions do not rely on `WriteFlags`.

victorporof: Removed, thanks!

victorporof:
Addressed comments.

victorporof:
Landed in feature branch `safe-mode`: https://github.com/mozilla/rkv/compare/safe-mode?expand=1

victorporof:
Closing in favor of https://github.com/mozilla/rkv/pull/159

PR #159, opened 2019-11-02
Merge safe mode to master and ship to crates.io

victorporof:
Safe mode is in a feature branch for now. Longer term, we want to start advertising RKV as Mozilla's simple and humane key-value storage solution, not just an LMDB wrapper.

There's a bunch more work that needs to happen, mostly with regards to documentation. This PR is now open as a followup of https://github.com/mozilla/rkv/pull/158


victorporof:
Locking this for now. We don't want any more changes to the safe mode branch at the moment.

PR #160, opened 2019-11-02
Safe mode arena allocation for databases

victorporof:
In https://github.com/mozilla/rkv/pull/158#issuecomment-548693019 I've explained why dropping the requirement for `Copy` and `Eq` traits was necessary for database instances.

Turns out I lied. We can use arena allocation and opaque keys as database handles. Doing so allows us to preserve the copy and equality trait requirements on "databases" (as users see them).

This patch does the following:
1. Re-enables copy and equality trait requirements on databases.
2. Returns database handles instead of database instances to consumers, in situations such as calling `open_db` or `create_db` on environments, or when dealing with transactions.
3. Derives copy and equality traits on single/multi database helpers.

This ends up simplifying things quite a lot, removes an entire class of errors, removes a layer of smart pointer indirection, and prevents all the copy/eq regressions from #158.

Depends on https://github.com/mozilla/rkv/pull/159.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ exclude = ["/tests/envs/*"]
 [features]
 default = ["with-safe-mode"]
 backtrace = ["failure/backtrace", "failure/std"]
-with-safe-mode = ["log", "uuid/v4", "uuid/serde", "serde/derive", "serde/rc"]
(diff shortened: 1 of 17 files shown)


victorporof:
As usual, the individual commits tell a story. Check them out individually.

PR #161, opened 2019-11-03
Cow snapshots instead of always cloning

victorporof:
To satisfy LMDB's semantics, querying or modifying individual transactions has certain behavior guarantees. Each transaction is a snapshot of the database; in other words, multiple read/write transactions are views into the database at the beginning of the transaction, altering or querying independent of each other.

To do this in safe mode, we create database snapshots when instantiating transactions (cloning on read). This is needlessly precautious, we can opt for cloning on write instead. This PR improves performance in case of large databases and the usual use of transactions.


src/backend/impl_safe/database.rs
```diff
@@ -70,31 +74,36 @@ impl Snapshot {
     }
 
     pub(crate) fn put_one(&mut self, key: &[u8], value: &[u8]) {
-        let values = self.map.entry(Box::from(key)).or_insert_with(BTreeSet::new);
+        let map = Arc::make_mut(&mut self.map);
```
ncloudioj: Would snapshotting in SafeMode support the multithreading scenario? If so, `Arc::make_mut()` might have cloned the underlying map if there are other readers holding the read reference of `self.map`. In which case, we might want to update the `self.map` with the new one.

victorporof: That's actually exactly what Arc::make_mut already does internally.

victorporof: At least that's what my interpretation of [this](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) seems to suggest.

PR #164, opened 2019-11-07
LMDB behavior parity

victorporof:
I've identified 3 additional situations where our safe mode backend implementation doesn't match LMDB's behavior, fixed in the following 3 individual commits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ arrayref = "0.3"
 bincode = "1.0"
 bitflags = "1"
 byteorder = "1"
-hashbrown = { version = "0.6", features = ["serde"] }
 id-arena = "2.2"
 lazy_static = "1.0"
 lmdb-rkv = "0.12.3"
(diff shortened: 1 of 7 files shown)


src/backend/impl_safe/cursor.rs
```diff
@@ -30,7 +30,7 @@ impl<'env> BackendRoCursor<'env> for RoCursorImpl<'env> {
     {
         // FIXME: Don't allocate.
         let key = key.as_ref().to_vec();
-        IterImpl(Box::new(self.0.iter().skip_while(move |&(k, _)| k != &key[..])))
+        IterImpl(Box::new(self.0.iter().skip_while(move |&(k, _)| k < key.as_slice())))
```
victorporof: This was making clippy complain unnecessarily. Rather than silencing it with a pragma, using `as_slice` made it shut up.

Cargo.toml
```diff
@@ -26,7 +26,6 @@ arrayref = "0.3"
 bincode = "1.0"
 bitflags = "1"
 byteorder = "1"
-hashbrown = { version = "0.6", optional = true }
```
victorporof: Also, this is unfortunate, but it the raw entry API doesn't exist for BTreeMap, which we need to use for sorted databases.

src/env.rs
```diff
@@ -304,24 +304,38 @@ where
     }
 }
 
+#[cfg(test)]
+macro_rules! make_check_rvk {
```
ncloudioj: s/make_check_rvk/make_check_rkv/ ?

PR #174, opened 2019-11-12
Bump version to 0.11.0

victorporof:


victorporof:
@ncloudioj Yup. Note that this PR isn't about merging to master, just making sure our vendoring process is actually able to pick up the changes. We'll still keep this in a feature branch for now.

PR #185, opened 2019-12-12
Bump UUID version to 0.8

rasviitanen:
We are tying out rkv for servo's IndexedDB.
Currently, servo depends on `UUID:0.8`. To avoid havning multiple versions of UUID, it would be great to bump up the UUID for `rkv` version to 0.8.


rasviitanen:
Oops, there was already an active PR for this issue.

PR #200, opened 2020-07-22
Implement a simple migrator between multiple backends

victorporof:
Needed to convert kvstore to use safe mode.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1597898 and dependent bugs.


src/store/single.rs
```diff
@@ -128,7 +128,7 @@ impl<'i, I> Iterator for Iter<'i, I>
 where
     I: BackendIter<'i>,
 {
-    type Item = Result<(&'i [u8], Option<Value<'i>>), StoreError>;
+    type Item = Result<(&'i [u8], Value<'i>), StoreError>;
```
ncloudioj, line 131: I like this change👍, it solves [this itch](https://github.com/mozilla/rkv/pull/35#issuecomment-395879259) for me :)

We need to at least bump the minor version as this breaks the old iterator API.

PR #201, opened 2020-07-23
Make migration easier for consumers

victorporof:

This makes it so that users don't need to manually open an old environment to migrate. Really they just want open their environment normally, and optionally populate it with with whatever was there to begin with.

This PR makes migration a one liner.

### DIFF
--- a/src/backend/impl_lmdb/environment.rs
+++ b/src/backend/impl_lmdb/environment.rs
@@ -10,7 +10,10 @@
 
 use std::{
     fs,
-    path::Path,
+    path::{
+        Path,
+        PathBuf,
+    },
 };
 
 use lmdb::Error as LmdbError;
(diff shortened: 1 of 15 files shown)


src/backend/impl_lmdb/environment.rs
```diff
@@ -189,6 +213,14 @@ impl<'e> BackendEnvironment<'e> for EnvironmentImpl {
...
     fn set_map_size(&self, size: usize) -> Result<(), Self::Error> {
-        self.0.set_map_size(size).map_err(ErrorImpl::LmdbError)
+        self.lmdbenv.set_map_size(size).map_err(ErrorImpl::LmdbError)
+    }
+
+    fn get_files_on_disk(&self) -> Vec<PathBuf> {
+        let mut db_filename = self.path.clone();
+        let mut lock_filename = self.path.clone();
+        db_filename.push("data.mdb");
+        lock_filename.push("lock.mdb");
```
ncloudioj, line 223: For the lmdb backend, users could opt out the lock file through `EnvironmentFlags::NO_LOCK`. Let's make sure this special case is covered.

victorporof: Nice catch, sounds good!

ncloudioj: Also, another option `NO_SUB_DIR` could change the default directory layout.

victorporof: The migrator says NO_SUB_DIR isn't supported yet, but indeed we shouldn't just assume other consumers won't use this methods directly. How do you suggest handling this case?

victorporof: I think for `NO_LOCK`, we should just not include the lock file in the output vector. For `NO_SUB_DIR`, we should return an error saying the directory layout is unknown.

ncloudioj: Those are rare cases, feel free to handle them in a followup.

victorporof: Well, should be an easy addition, thanks for spotting these! I'll go ahead with https://github.com/mozilla/rkv/pull/201#discussion_r459695976

victorporof:
Addressed comments.

PR #203, opened 2020-07-28
Ignore file invalid errors when migrating

victorporof:
Really there's not much that can be done in this situation. The "easy mode" auto migration should handle this automatically, similarly to how other IoError and UnsuitableEnvironmentPath errors are ignored. The non-auto migration is still available for manual/complex handling of such problems.

### DIFF
--- a/src/migrator.rs
+++ b/src/migrator.rs
@@ -18,6 +18,19 @@
 //!
 //! The destination environment should be empty of data, otherwise an error is returned.
 //!
+//! There are 3 versions of the migration methods:
+//! * `migrate_<src>_to_<dst>`, where `<src>` and `<dst>` are the source and destination
+//!   environment types. You're responsive with opening both these environments, handling
+//!   all errors, and performing any cleanup if necessary.
+//! * `open_and_migrate_<src>_to_<dst>`, which is similar the the above, but automatically
+//!   attempts to open the source environment and delete all of its supporting files if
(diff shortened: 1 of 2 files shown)


src/migrator.rs
```diff
@@ -88,8 +88,8 @@ macro_rules! fn_migrator {
 
     ($migrate:tt, $name:tt, $builder:tt, $src_env:ty, $dst_env:ty) => {
         /// Same as the other migration methods, but automatically attempts to open the source
-        /// environment, ignores it if it doesn't exist or if it's empty, and finally attempts to
-        /// delete all of its supporting files.
+        /// environment, ignores it if it's invalid, if it doesn't exist or if it's empty.
```
ncloudioj, line 91: Nit: s/`..., ignores it if it's invalid, if it doesn't exist or if it's empty.`/`..., ignores the migration if the source environment is invalid, inexistent, or empty.`/

Since this migrator will delete all the supporting files, silently ignoring errors without notifying the caller might sound surprising to them. Let's add a note to the migration document so that the consumers can decide which migrator is the best suit for them.

victorporof: Note: it won't delete the files if they're invalid (corrupted, or unrelated). This is intentional.

I'll expand on picking the right migrator in the comments.

victorporof:
I've added a commit that addresses comments and makes error handling less surprising. It also documents which migration method is suitable for which kind of task to make it easier to pick between them.

PR #206, opened 2020-10-09
Migration updates

victorporof:
This PR does 2 things:

1. Fixes bug https://bugzilla.mozilla.org/show_bug.cgi?id=1663856

This is a fun one. If old migrated data isn't deleted (e.g. when there's multiple environments open at the same path and deleting the files would be invalid while another environment is still open), it's possible that databases with new a different storage drivers (e.g. safe-mode) continue to be created and end up populated at runtime. Next time migration is attempted, migration doesn't fail silently anymore (since the destination is not empty). See [comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1663856#c5)).

I've added some tests for this and now handling the `DestinationNotEmpty` situations. In every situation, the consumer code that's using this helper:
- Is not concerned by throwing away old corrupted data and starting fresh with a new store.
- Will never want to overwrite data in the new store from the old corrupted store.

2. Prevent potential deadlocks in safe mode environments when opening dbs.

Two rwlocks is one too many. This is theoretical and nobody's encountered this yet afaik, but is caught by tsan: https://bugzilla.mozilla.org/show_bug.cgi?id=1606804. Switching to a single rwlock _should_ fix it I think, and it shouldn't hurt in general.

### DIFF
--- a/src/backend/impl_safe/environment.rs
+++ b/src/backend/impl_safe/environment.rs
@@ -12,6 +12,7 @@ use std::{
     borrow::Cow,
     collections::HashMap,
     fs,
+    ops::DerefMut,
     path::{
         Path,
         PathBuf,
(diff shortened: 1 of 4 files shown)


ncloudioj (approved):
LGTM!

Wow, such a tricky bug.

While this fixes the issue, there are still some questions worth further considering.
* If the consumer somehow keeps both the old and the new environments, the same migration will be attempted again upon the next use, right?
* I wonder if we can do better in this case instead of simply aborting the migration? For instance, how about the migrator tried its best to merge the two envs if they both exist and non-empty, and let the consumer decide upfront (e.g. as a function argument) which one to keep when conflicts detected.

@victorporof What do you think?

victorporof:
You're correct, migration would be attempted again next time, but skipped because the destination is not empty. I definitely agree that merging would be quite elegant. I'd be very happy to get a PR for this, but unlikely to implement this right now for now given our particular use cases where we don't care that much about the old data (and merging is a tricky problem in general!).

Thanks for the quick review.

PR #207, opened 2020-10-09
Add a way to discard corrupted databases on load

victorporof:
Right now, if database files are corrupted, opening an environment at that path will fail with `FileInvalid` (every time with SafeMode, and only _sometimes_ with LMDB unless corruption is extreme in which case we just crash).

The only way for users to deal this right now is to manually delete the files before attempting to reopen –– which is also incorrect if there's lingering environments. This is also very inelegant when dealing with multiple possible backends, where users might not know exactly which files to delete.

I've recently added `try_close_and_delete` to Managers, and `close_and_delete` to Environments, which delete the files on close. This is only useful though if you've already successfully opened an environment at least once with the former, or already have direct ownership of an environment with the latter, and can therefore guarantee that there's no other environments open in other threads for example.

This PR makes the following changes:
* Adds a `set_discard_if_corrupted` method to environment builders to actually deal with this problem.
* Turns Environments `close_and_delete(self)` into `close(self, options: CloseOptions)`, that makes deleting files optional. This is useful for testing, but I can imagine it being useful in other situations as well.
* Adds a specific class of errors called `CloseError` which only happen when closing, to distinguish them from `StoreError`s. This helps isolating errors coming from a particular part of a migration process for example, where IOError could mean "failed to open", as well as "failed to delete while closing".

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -30,13 +30,17 @@ use crate::{
         BackendRwCursorTransaction,
         SafeModeError,
     },
-    error::StoreError,
+    error::{
+        CloseError,
+        StoreError,
+    },
     readwrite::{
         Reader,
         Writer,
     },
     store::{
         single::SingleStore,
+        CloseOptions,
         Options as StoreOptions,
     },
 };
@@ -310,15 +314,16 @@ where
         self.env.set_map_size(size).map_err(Into::into)
     }
 
-    /// Closes this environment and deletes all its files from disk. Doesn't delete the
-    /// folder used when opening the environment.
-    pub fn close_and_delete(self) -> Result<(), StoreError> {
+    /// Closes this environment and optionally deletes all its files from disk. Doesn't
+    /// delete the folder used when opening the environment.
+    pub fn close(self, options: CloseOptions) -> Result<(), CloseError> {
         let files = self.env.get_files_on_disk();
-        self.sync(true)?;
         drop(self);
 
-        for file in files {
-            fs::remove_file(file)?;
+        if options.delete {
+            for file in files {
+                fs::remove_file(file)?;
+            }
         }
 
         Ok(())
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -34,12 +34,17 @@ use crate::{
         LmdbEnvironment,
         SafeModeEnvironment,
     },
-    error::StoreError,
+    error::{
+        CloseError,
+        StoreError,
+    },
     helpers::canonicalize_path,
+    store::CloseOptions,
     Rkv,
 };
 
 type Result<T> = result::Result<T, StoreError>;
+type CloseResult<T> = result::Result<T, CloseError>;
 type SharedRkv<E> = Arc<RwLock<Rkv<E>>>;
 
 lazy_static! {
@@ -146,10 +151,9 @@ where
         })
     }
 
-    /// Tries to close the specified environment and delete all its files from disk.
-    /// Doesn't delete the folder used when opening the environment.
-    /// This will only work if there's no other users of this environment.
-    pub fn try_close_and_delete<'p, P>(&mut self, path: P) -> Result<()>
+    /// Tries to close the specified environment.
+    /// Returns an error when other users of this environment still exist.
+    pub fn try_close<'p, P>(&mut self, path: P, options: CloseOptions) -> CloseResult<()>
     where
         P: Into<&'p Path>,
     {
@@ -159,16 +163,14 @@ where
             canonicalize_path(path)?
         };
         match self.environments.entry(canonical) {
-            Entry::Vacant(_) => {}, // noop
+            Entry::Vacant(_) => Ok(()),
+            Entry::Occupied(e) if Arc::strong_count(e.get()) > 1 => Err(CloseError::EnvironmentStillOpen),
             Entry::Occupied(e) => {
-                if Arc::strong_count(e.get()) == 1 {
-                    if let Ok(env) = Arc::try_unwrap(e.remove()) {
-                        env.into_inner()?.close_and_delete()?;
-                    }
-                }
+                let env = Arc::try_unwrap(e.remove()).map_err(|_| CloseError::UnknownEnvironmentStillOpen)?;
+                env.into_inner()?.close(options)?;
+                Ok(())
             },
         }
-        Ok(())
     }
 }
 
--- a/tests/manager.rs
+++ b/tests/manager.rs
@@ -17,12 +17,16 @@ use tempfile::Builder;
 
 use rkv::{
     backend::{
+        BackendEnvironmentBuilder,
         Lmdb,
         LmdbEnvironment,
         SafeMode,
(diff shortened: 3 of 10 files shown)


tests/manager.rs
```diff
@@ -134,3 +138,120 @@ fn test_same_with_capacity_safe() {
...
+    manager.try_close(root.path(), CloseOptions::default()).expect("closed without deleting");
+    assert!(manager.get(root.path()).expect("success").is_none());
+
+    // Recreating environment fails.
+    manager.get_or_create(root.path(), Rkv::new::<SafeMode>).expect_err("not created");
+    assert!(manager.get(root.path()).expect("success").is_none());
+
+    // But we can use a builder and pass `discard_if_corrupted` to deal with it.
+    let mut builder = Rkv::environment_builder::<SafeMode>();
+    builder.set_discard_if_corrupted(true);
```
victorporof, line 185: This is the secret sauce.

src/manager.rs
```diff
@@ -146,10 +151,9 @@ where
         })
     }
 
-    /// Tries to close the specified environment and delete all its files from disk.
-    /// Doesn't delete the folder used when opening the environment.
-    /// This will only work if there's no other users of this environment.
-    pub fn try_close_and_delete<'p, P>(&mut self, path: P) -> Result<()>
+    /// Tries to close the specified environment.
+    /// Returns an error when other other users of this environment still exist.
```
victorporof, line 155: This used to fail silently if other environments existed. Now it returns an error instead.

tests/manager.rs
```diff
@@ -134,3 +138,120 @@ fn test_same_with_capacity_safe() {
...
+    // Verify it was flushed to disk.
+    let mut safebin = root.path().to_path_buf();
+    safebin.push("data.safe.bin");
+    assert!(safebin.exists());
+
+    // Oops, corruption.
+    fs::write(&safebin, "bogus").expect("dbfile corrupted");
+
+    // Reading still works. Magic.
+    let store = env.open_single("store", StoreOptions::default()).expect("opened");
```
victorporof, line 222: Magic.

ncloudioj:
Just took a glimpse at the code change, this features looks nice and neat to me. I'd like to take a closer look next week as the actual review if you don't mind.

Also, this makes me wonder that perhaps the different corruption resistance level is the biggest distinction between those two backends in rkv. To my understanding, LMDB is generally corruption-tolerant if not -free unless you opt-in the `MDB_WRITEMAP` mode, which rkv doesn't expose. Its shadow-paging based MVCC architecture can even survive from the system crushes (i.e partial flush) at the cost of data loss but not corruption. Whereas for SafeMode backend, I don't think it can achieve the same level of resistance as any interruption during the environment flush could lead to a corrupted file. So I'd argue that it's more likely to run into this kind of trouble with the SafeMode.

While I am not sure how often this would happen in the wild, shall we consider raising rkv users' awareness (maybe by documenting?) about this difference? I believe that will be useful for them to decide which backend to use. More importantly, they will be prepared to deal with it and to leverage features like this when using SafeMode in their code.

@victorporof What do you think?

src/env.rs
```diff
@@ -310,15 +314,16 @@ where
...
     }
 
-    /// Closes this environment and deletes all its files from disk. Doesn't delete the
-    /// folder used when opening the environment.
-    pub fn close_and_delete(self) -> Result<(), StoreError> {
+    /// Closes this environment and optionally deletes all its files from disk. Doesn't
+    /// delete the folder used when opening the environment.
+    pub fn close(self, options: CloseOptions) -> Result<(), CloseError> {
         let files = self.env.get_files_on_disk();
-        self.sync(true)?;
```
ncloudioj, line 317: Why the force sync is no longer needed here for both backends? If it's still useful to the consumers, we can add it to `CloseOptions` as another option.

src/manager.rs
ncloudioj, line 155: s/other other/other/

tests/manager.rs
```diff
@@ -134,3 +138,120 @@ fn test_same_with_capacity_safe() {
     let fetched_arc = manager.get(p).expect("success").expect("existed");
     assert!(Arc::ptr_eq(&created_arc, &fetched_arc));
 }
+
+/// Some storage drivers are able to recover when the database is corrupted at runtime.
+/// Test how these managers can discard corrupted databases and re-open.
```
ncloudioj, line 143: Nit: this test is to verify that a corrupted database can be opened with `discard_if_corrupted`, but the comment (which is exactly the same as the one of the next test) sounds more about being able to recover from the db corruption, which is a bit misleading IMO.

tests/manager.rs
ncloudioj: It's indeed magic!! How did it manage to survive from the previous corruption? You didn't even call `set_discard_if_corrupted()` here.

ncloudioj (approved):
LGTM. r+ with a few nits.

ncloudioj:
Just a reminder, this needs a version bump due to the API changes.

tests/manager.rs
victorporof: Database is kept in memory for safe-mode.

src/env.rs
victorporof: It was never needed to begin with! I initially added it because it seemed like a good default, but we never open LMDB with MDB_NOSYNC, and in the case of safe-mode it's a noop.

tests/manager.rs
victorporof: Reworded.

victorporof:
> Just took a glimpse at the code change, this features looks nice and neat to me. I'd like to take a closer look next week as the actual review if you don't mind.
> 
> Also, this makes me wonder that perhaps the different corruption resistance level is the biggest distinction between those two backends in rkv. To my understanding, LMDB is generally corruption-tolerant if not -free unless you opt-in the `MDB_WRITEMAP` mode, which rkv doesn't expose. Its shadow-paging based MVCC architecture can even survive from the system crushes (i.e partial flush) at the cost of data loss but not corruption. Whereas for SafeMode backend, I don't think it can achieve the same level of resistance as any interruption during the environment flush could lead to a corrupted file. So I'd argue that it's more likely to run into this kind of trouble with the SafeMode.
> 
> While I am not sure how often this would happen in the wild, shall we consider raising rkv users' awareness (maybe by documenting?) about this difference? I believe that will be useful for them to decide which backend to use. More importantly, they will be prepared to deal with it and to leverage features like this when using SafeMode in their code.
> 
> @victorporof What do you think?

The two different backers indeed have very different theoretical corruption tolerances. In the wild though (for our uses), the corruption has always happened on the LMDB side on Window 7 machines. These PRs are actually here to deal with that problem, where data is unrecoverable and messes up migrations to safe-mode. There's been no instances of corruption with safe-mode yet, however the actual real-world tolerance remains to be proven in practice as more and more stores migrate to using it.

victorporof:
Addressed comments.

PR #209, opened 2020-11-09
Update dependencies to minimal supported versions

badboy:
I am trying to build Glean using minimal versions (`cargo +nightly update -Z minimal-versions`) and this is another dependency that breaks.

ordered-float dropped the `unreachable` dependency in 1.0.1, which makes it work again.


badboy:
also requires https://github.com/mozilla/rkv/pull/209

victorporof:
@badboy Uh, this is #209.

badboy:
I .... am not sure which PR I then menat.

ncloudioj:
This looks good to me.

@victorporof Perhaps we can bundle this with the other one in #210, and vendor them to m-c?

PR #210, opened 2020-12-05
Replace failure with thiserror

upsuper:
[`failure`](https://github.com/rust-lang-nursery/failure) has been deprecated in favor of `thiserror`. This PR replaces it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ repository = "https://github.com/mozilla/rkv"
 version = "0.16.0"
 
 [features]
-backtrace = ["failure/backtrace", "failure/std"]
 db-dup-sort = []
 db-int-key = []
 default = ["db-dup-sort", "db-int-key"]
@@ -41,15 +40,9 @@ ordered-float = "1.0"
 paste = "0.1"
 serde = {version = "1.0", features = ["derive", "rc"]}
 serde_derive = "1.0"
+thiserror = "1.0"
 url = "2.0"
 uuid = "0.8"
-
-# Get rid of failure's dependency on backtrace. Eventually
-# backtrace will move into Rust core, but we don't need it here.
-[dependencies.failure]
-default_features = false
-features = ["derive"]
-version = "0.1"
 
 [dev-dependencies]
 byteorder = "1"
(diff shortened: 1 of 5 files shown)


victorporof:
I wonder if this conflicts with usage in m-c at all.

upsuper:
I'm going to update m-c as well.

victorporof:
@upsuper Do you have a link to a bug in m-c?

upsuper:
Nope. I'll create one when I have the patch :)

I don't think this would affect downstream crates, as `Fail` is implemented for all `Error` types. The only meaningful difference is that backtrace support is dropped, which m-c doesn't use anyway.

upsuper:
Meta bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1681898

upsuper:
Oh actually there is a security advisory about the deprecation of failure: https://rustsec.org/advisories/RUSTSEC-2020-0036.html

upsuper:
Based on experience on agashlin/comedy-rs#13, this would require a minor version bump, because `failure` without `std` feature doesn't really have `impl Fail` for error.

upsuper:
Hey @victorporof, what do you think about this?

After [removing `failure` from `bits_client`](https://bugzilla.mozilla.org/show_bug.cgi?id=1684238), dependents of rkv and crates of lucet sandbox compilers are the last crates which are still using `failure`. I've contacted maintainer of lucet sandbox compilers, and from January next year, they are going to work on catching up with their upstream which has dropped `failure` dependencies for long. By then dependents of rkv would be the only reason we can't remove dependency of `failure` in m-c.

If you can merge this and release a new version, I can raise patch to update m-c to vendor the new version and fix up things.

upsuper:
Ping for this.

ncloudioj:
@upsuper My apologies for the late response.

Looking into your patch now.

ncloudioj (approved):
LGTM! @upsuper Thanks for the patch! 

@victorporof Could you bump the version (minor) and publish it so that @upsuper can proceed with the rest of the work in m-c?

upsuper:
Thanks! I'll look into this over this weekend.

victorporof:
Done.

PR #218, opened 2022-06-01
Replace travis CI/AppVeyor with CircleCI

badboy:
Note:
This is currently a very reduced feature set.

* Only stable
* Only Linux

---

Note that while this removes the old config, the services are still set up to do work.
Trying to figure out who can help with that.


saschanaz:
I'd still prefer GitHub Actions unless there's a good reason to prefer CircleCI, 1) just because I'm more familiar with GHA 2) that way we won't need to go another website and the permission to modify settings 👀

badboy:
I'm ok with GhA. We don't need anything special, so all already-allowed actions should be good enough for us.
I would leave the setup to you then though.

PR #221, opened 2022-06-07
`cargo clippy --fix`

saschanaz:
Only trivial fixes here.

See https://github.com/mozilla/rkv/runs/6781744019?check_suite_focus=true for the full log.


badboy (approved):

saschanaz:
#223 did this 🎉

PR #223, opened 2022-07-20
Put LMDB support behind a feature

badboy:
This would disable the LMDB backend by default.
Note that for some time Glean will still enable that ([bug 1772148](https://bugzilla.mozilla.org/show_bug.cgi?id=1772148)), though disabling that in m-c (and on Android) might come sooner than on our iOS builds.

In the future then fully removing it is as easy as removing the gated code.


saschanaz (changes requested):
Quite a lot of compiler errors, do you think you can get some time to fix it? Otherwise I can take this.

badboy:
> Quite a lot of compiler errors, do you think you can get some time to fix it? Otherwise I can take this.

"worked on my machine" -- hehe. Yes, I'll take a look.

badboy:
All kinds of other weird errors now. I'll look into that tomorrow

badboy:
Now this is starting to look better.
But it required a couple of unrelated changes to get CI to work again, while avoiding a MSRV bump.
I plan to do some cleanup (and really an MSRV bump) as a followup.

src/backend/impl_lmdb/arch_migrator.rs
```diff
@@ -7,6 +7,8 @@
 // under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 // CONDITIONS OF ANY KIND, either express or implied. See the License for the
 // specific language governing permissions and limitations under the License.
+#![allow(dead_code)] // lots of unused struct members
+#![allow(clippy::upper_case_acronyms)] // things like BRANCH
```
saschanaz, line 11: Just one thing, could you add TODO or XXX to indicate these are not really intentional but just temporary workaround? (Unless you intend to do the followup today 😜)

saschanaz (approved):
Thank you, everything looks good! I super appreciate the commit structure.

src/backend/impl_lmdb/arch_migrator.rs
badboy: guess these two specifically won't get addressed anway. It's LMDB code, that goes away when we drop LMDB support fully.

PR #224, opened 2022-08-12
Follow the default behavior of rustfmt

saschanaz:
I don't think it's a lot worth to keep unstable formatting options in a non-experimental public library. Maintaining `KNOWN_TO_HAVE_RUSTFMT` solely for those options is not ideal at all, so perhaps we should just follow the default behavior unless there's a big reason not to do that.

What do you think?


.rustfmt.toml
```diff
@@ -1,18 +0,0 @@
-match_block_trailing_comma = true
-max_width = 120
```
saschanaz, line 2: This is especially weird, this is a Mozilla library and still doesn't follow what mozilla-central does (the 80-char default width from clang-format) 🤔

badboy (approved):
I'm all for this. Defaults all the way!

Issue #222, opened 2022-06-08
Remove LMDB backend


saschanaz:
I think we should still keep the migrator for some time.

saschanaz:
Not sure what to do with bin/dump and bin/rand, it's currently only for LMDB backend. Any actual user for those? 🤔

rand should be able to support safemode but dump uses migrator that only LMDB has.

badboy:
> Not sure what to do with bin/dump and bin/rand, it's currently only for LMDB backend. Any actual user for those? 🤔
> 
> rand should be able to support safemode but dump uses migrator that only LMDB has.

These always felt more like examples anyway. I'm fine if they are gone.

PR #225, opened 2022-08-12
Bump MSRV to Rust 1.58.0

badboy:


saschanaz (approved):
Approving, but it might be a good chance to replace `.panic("{}", e)` to `.panic_any(e)`. Clippy suggested the latter but it caused errors in #221 on the old version. How about it?

badboy:
> Approving, but it might be a good chance to replace `.panic("{}", e)` to `.panic_any(e)`. Clippy suggested the latter but it caused errors in #221 on the old version. How about it?

rustc warns about `panic!(err);` being deprecated and either `panic!("{}", e); or `panic_any!(e)` to replace it.
We had `panic!(err)` in a couple of places and I fixed those in #222 to use `panic!("{}", e)`.
They are only in tests:

```
❯ rg 'panic.\("\{\}", err'
src/backend/impl_lmdb/arch_migrator.rs
773:                Err(err) => panic!("{}", err),
775:                    Err(err) => panic!("{}", err),

tests/env-lmdb.rs
1519:                Err(err) => panic!("{}", err),

tests/env-safe.rs
1242:                Err(err) => panic!("{}", err),
```

No point in touching those.

saschanaz:
Ah yes, I somehow thought I saw more but I'm wrong 😬, sorry. Thanks for correcting me.

PR #230, opened 2022-11-15
Allow bitflags 1.2+

badboy:
Not sure why this was restricted, but unless we losen this dependency range we cannot get it into m-c.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ with-fuzzer-no-link = ["lmdb", "lmdb-rkv/with-fuzzer-no-link"]
 [dependencies]
 arrayref = "0.3"
 bincode = "1.0"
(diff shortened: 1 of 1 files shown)


saschanaz:
That was https://github.com/mozilla/rkv/pull/220#discussion_r891734736. https://github.com/mozilla/rkv/pull/225 changed the MSRV so we should be okay now.

saschanaz (approved):
Approving, but do we want to bump it to 1.3?

badboy:
> Approving, but do we want to bump it to 1.3? Dependabot will nag us very soon anyway.
> 
> Edit: Oh wait, we don't have Dependabot set.

It also doesn't matter, because this just means `^1.2` so, everything `>= 1.2` and `<2`

PR #231, opened 2022-11-24
Write database to temp file first, then atomically move it

badboy:
This ensures that _some_ data is still around in the case that the write gets interrupted early.
This currently can cause data loss with the whole database getting wiped.

### DIFF
--- a/src/backend/impl_safe/environment.rs
+++ b/src/backend/impl_safe/environment.rs
@@ -217,7 +217,13 @@ impl EnvironmentImpl {
         if fs::metadata(&path)?.is_dir() {
             path.to_mut().push(DEFAULT_DB_FILENAME);
         };
-        fs::write(&path, self.serialize()?)?;
+
+        // Write to a temp file first.
+        let tmp_path = path.with_extension("tmp");
+        fs::write(&tmp_path, self.serialize()?)?;
+
+        // Atomically move that file to the database file.
+        fs::rename(tmp_path, path)?;
         Ok(())
     }
 


saschanaz (approved):
LGTM with a question

src/backend/impl_safe/environment.rs
```diff
@@ -217,7 +217,13 @@ impl EnvironmentImpl {
...
             path.to_mut().push(DEFAULT_DB_FILENAME);
         };
-        fs::write(&path, self.serialize()?)?;
+
+        // Write to a temp file first.
+        let tmp_path = path.with_extension("tmp");
+        fs::write(&tmp_path, self.serialize()?)?;
+
+        // Atomically move that file to the database file.
+        fs::rename(tmp_path, path)?;
```
saschanaz, line 226: Forgot to submit.

Should we do any cleanup when this fails? We may get some redundant tmp files in rare cases.

badboy: Hm, the worst that happens is that we keep a `.tmp` file until the next write, when it will just be overwritten.
At least for Glean databases are small in size and so the potential doubling of space used here doesn't really matter.
The major case would be when we can't actually write to `path`. But then why would we be able to write to `tmp_path` to begin with. In this case it might be interesting to keep the `tmp_path` around and be able to recover it later.

badboy: That plus the annoying additional logic for the error case made me not do it.

saschanaz: Ah, makes sense. Thanks for the explanation.

PR #235, opened 2023-03-24
Introduce a recovery strategy as a replacement for set_discard_if_corrupted

### LINKED ISSUE #234: Replace set_discard_if_corrupted with a set_corruption_recovery_strategy
```rust
enum RecoveryStrategy {
    /// Discard the corrupted data and start with an empty database
    Discard,
    /// Move the corrupted data file to `$file.corrupt` and start with an empty database
    Rename,
    /// Bubble up the error on detecting a corrupted data file
    Error,
}
```

See https://bugzilla.mozilla.org/show_bug.cgi?id=1645907#c11

badboy:
Fixes #234

This is a breaking change because of the now-changed public API

### DIFF
--- a/src/backend/impl_safe/environment.rs
+++ b/src/backend/impl_safe/environment.rs
@@ -24,9 +24,11 @@ use super::{
     database::Database, DatabaseFlagsImpl, DatabaseImpl, EnvironmentFlagsImpl, ErrorImpl, InfoImpl,
     RoTransactionImpl, RwTransactionImpl, StatImpl,
 };
+use crate::backend::common::RecoveryStrategy;
 use crate::backend::traits::{BackendEnvironment, BackendEnvironmentBuilder};
 
 const DEFAULT_DB_FILENAME: &str = "data.safe.bin";
+const DEFAULT_CORRUPT_DB_EXTENSION: &str = "bin.corrupt";
 
 type DatabaseArena = Arena<Database>;
 type DatabaseNameMap = HashMap<Option<String>, DatabaseImpl>;
@@ -38,7 +40,7 @@ pub struct EnvironmentBuilderImpl {
     max_dbs: Option<usize>,
     map_size: Option<usize>,
     make_dir_if_needed: bool,
-    discard_if_corrupted: bool,
+    corruption_recovery_strategy: RecoveryStrategy,
 }
 
 impl<'b> BackendEnvironmentBuilder<'b> for EnvironmentBuilderImpl {
@@ -53,7 +55,7 @@ impl<'b> BackendEnvironmentBuilder<'b> for EnvironmentBuilderImpl {
             max_dbs: None,
             map_size: None,
             make_dir_if_needed: false,
-            discard_if_corrupted: false,
+            corruption_recovery_strategy: RecoveryStrategy::Error,
         }
     }
 
@@ -85,8 +87,8 @@ impl<'b> BackendEnvironmentBuilder<'b> for EnvironmentBuilderImpl {
         self
     }
 
-    fn set_discard_if_corrupted(&mut self, discard_if_corrupted: bool) -> &mut Self {
-        self.discard_if_corrupted = discard_if_corrupted;
+    fn set_corruption_recovery_strategy(&mut self, strategy: RecoveryStrategy) -> &mut Self {
+        self.corruption_recovery_strategy = strategy;
         self
     }
 
@@ -106,7 +108,7 @@ impl<'b> BackendEnvironmentBuilder<'b> for EnvironmentBuilderImpl {
             self.max_dbs,
             self.map_size,
         )?;
-        env.read_from_disk(self.discard_if_corrupted)?;
+        env.read_from_disk(self.corruption_recovery_strategy)?;
         Ok(env)
     }
 }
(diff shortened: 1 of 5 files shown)


badboy:
SIGABRTs for the mac tests. yey, gonna need to look into that.

badboy:
SIGABRTs gone now, with latest Rust (1.68.2, released hours ago). I suspect some resource exhaustion, which happens to be gone now? 🤷

saschanaz:
(Closing and reopening to retrigger the CI)

tests/manager.rs
```diff
@@ -378,3 +378,97 @@ fn test_safe_mode_corrupt_while_open_2() {
...
+fn test_safe_mode_corrupt_while_open_3() {
+    type Manager = rkv::Manager<SafeModeEnvironment>;
+
+    let root = Builder::new()
+        .prefix("test_safe_mode_corrupt_while_open_3")
+        .tempdir()
+        .expect("tempdir");
+    fs::create_dir_all(root.path()).expect("dir created");
+
+    // Verify it was flushed to disk.
```
saschanaz, line 394: I see you copypasted this line from `test_safe_mode_corrupt_while_open_1`, but this comment doesn't seem to make sense when the `exists()` call now happens after the write call.

tests/manager.rs
```diff
@@ -378,3 +378,97 @@ fn test_safe_mode_corrupt_while_open_2() {
...
+
+    // But we can use a builder and pass `RecoveryStrategy::Rename` to deal with it.
+    let mut builder = Rkv::environment_builder::<SafeMode>();
+    builder.set_corruption_recovery_strategy(RecoveryStrategy::Rename);
+    manager
+        .get_or_create_from_builder(root.path(), builder, Rkv::from_builder::<SafeMode>)
+        .expect("created");
+    assert!(manager.get(root.path()).expect("success").is_some());
+
+    // Database file was moved out of the way.
```
saschanaz, line 419: Perhaps put the message into the assert call? (Also for similar cases below)

src/manager.rs
```diff
@@ -244,4 +245,46 @@ mod tests {
             .expect("success");
         assert!(!Arc::ptr_eq(&path2_arc, &arc));
     }
+
+    /// Test that one can mutate managed Rkv instances in surprising ways.
```
saschanaz, line 249: Might be nice to mention https://github.com/mozilla/rkv/pull/130 (or a corresponding issue, but I don't see one)

This doesn't seem to be directly related to the issue, did you somehow suffered from this behavior while writing the patch? 🙂

saschanaz (approved):
Looks good. The build failures are because of `log` doing MSRV bump without a corresponding major version bump.

saschanaz:
BTW, sorry for the huge delay, this just slipped. I hope there's a central Mozilla dashboard to manage both Phabricator and GitHub 😅

Or at least a separate GitHub dashboard for my paid job...

src/backend/impl_safe/environment.rs
```diff
@@ -152,16 +154,32 @@ impl EnvironmentImpl {
...
+        let bytes = fs::read(path)?;
+
+        match Self::deserialize(&bytes) {
+            Ok((arena, name_map)) => Ok((arena, name_map)),
+            Err(err) => match strategy {
+                RecoveryStrategy::Error => Err(err),
+                RecoveryStrategy::Discard => Ok((DatabaseArena::new(), HashMap::new())),
+                RecoveryStrategy::Rename => {
+                    let corrupted_path = path.with_extension(DEFAULT_CORRUPT_DB_EXTENSION);
+                    fs::rename(path, corrupted_path)?;
```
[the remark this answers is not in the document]
badboy: I guess the assumption was that one deals with the `$db.corrupt` file soon, before hitting another problem. But I can see that that's not always possible.
If we use timestamps rkv-users need to make sure to actually clean up to avoid piling up old corrupted files.

src/manager.rs
badboy: Wow, it's been some time :D -- I think so? I don't remember how I came to write this.

badboy:
I addresses the first few things, but I need to think about the filename part a bit more.

src/backend/impl_safe/environment.rs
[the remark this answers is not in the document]
saschanaz: By using Rename I'd expect the user to deal with the files, indeed. I wonder what they can really do with it if rkv can't do anything, though.

src/manager.rs
badboy: Ok, I really have no idea what exactly I wanted to show with this test. Maybe that's indeed #130? 
Let me take this commit out of this PR though.

src/backend/impl_safe/environment.rs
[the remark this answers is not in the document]
badboy: For debugging purporses someone could manually explore that file and see if they are able to detect why the file got corrupted.

I thought about the `.corrupt` vs `.$timestamp` option again and I would say we should stay with `.corrupt` for now:
1. It's what's been discussed a while back
2. It avoids piling up files over files
3. The logic to find it is kept simple
4. We still have the option in the future to change it. Let's see if it's even gonna be adopted inside m-c

PR #236, opened 2023-03-28
CI: Only run on pushes to main & cancel stale executions

badboy:


.github/workflows/ci.yml
```diff
@@ -1,9 +1,17 @@
...
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches: main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
saschanaz, line 10: Huh, this is new to me. Cool.

saschanaz (approved):

saschanaz:
So the SIGABRT thing is some Rust compiler regression?

saschanaz:
For now I retriggered the failing job. I also wonder we have any stress-testing test here, in that case perhaps it's better to split it somehow to a separate job. Very weird to hit the resource limit if we don't have one.

badboy:
> For now I retriggered the failing job. I also wonder we have any stress-testing test here, in that case perhaps it's better to split it somehow to a separate job. Very weird to hit the resource limit if we don't have one.

Again: that was just my suspicion. I can't say for certain that this is the cause.

saschanaz:
Retriggering made it green!

PR #239, opened 2023-10-31
Remove lmdb

### LINKED ISSUE #222: Remove LMDB backend

saschanaz:
Closes #222


tests/env-lmdb.rs
```diff
@@ -1,1647 +0,0 @@
-// Copyright 2018-2019 Mozilla
```
saschanaz, line 1: Note for me: revive this

badboy (approved):

